PR #16, opened 2018-06-05
Create linq test. #9

dalikkk:


linq/UnitTest1.cs
```diff
@@ -0,0 +1,157 @@
...
+                {
+                    Console.WriteLine(c);
+                }
+    
+    
+                var times = 0;
+                var shuffle = startingDeck;
+                do
+                {
+                    shuffle = shuffle.Skip(26).InterleaveSequenceWith(shuffle.Take(26)).ToArray();
```
RheaAyase: You can actually write an assert here as well, we know what should the resulting sequence look like. It would begin with the 27th element, then 1st, then 28th, then 2nd, then 29th, then 3rd, etc etc...
I'd make it a for loop to check whether the `shuffle[i]` is equal to...
```
if( i % 2 == 0 ) assert( secondhalf[(int)(i/2)] )
else assert( firsthalf[Math.Ceil(i/2f)] )
```
(second half being the `shuffle.Skip(26)` and the first half is `shuffle.Take(26)` - you need to create temporary arrays similarly as it is done with shuffle - `toarray`)

linq/UnitTest1.cs
```diff
@@ -69,24 +69,17 @@ public static void v()
...
-                    foreach (var c in shuffle)
-                        Console.WriteLine(c);
-    
-                    Console.WriteLine();
-                    */
+                    var result = shuffle.Skip(26).InterleaveSequenceWith(shuffle.Take(26)).ToArray();
+                    if( times % 2 == 0 )
+                        Assert.Equal( shuffle[times/2 + result.Length/2], result[times] );
+                    else Assert.Equal( shuffle[times/2], result[times] );
+                    shuffle = result;
```
RheaAyase: Not what I meant, it **has to** be in it's own loop, you should **not** use `times` variable in the assert - just like the console.writeline had it's own loop every single time, this has to have it's own loop as well, to pretty much replace it...

PR #19, opened 2018-06-19
Add ASP.NET Core patch tests

tmds:
Verifies each sdk uses the latest ASP.NET Core App/All package.
Verifies we can override the version used by the SDK using an envvar.

### DIFF
--- a/aspnetpatch-21/test.json
+++ b/aspnetpatch-21/test.json
@@ -1,0 +1,7 @@
+{
+  "name": "aspnetpatch-21",
+  "version": "2.1",
+  "versionSpecific": true,
+  "type": "bash",
+  "cleanup": true
+}
--- a/aspnetpatch-21/test.sh
+++ b/aspnetpatch-21/test.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+set -e
+
+latest_aspnet_package=2.1.1
+sdks=$(dotnet --list-sdks | awk '{print $1}')
+test_folder=testapp
+
+function create_project_for_sdk_and_package()
+{
+  local sdk=$1
+  local package=$2
+
+  # create folder
+  rm -rf $test_folder
+  mkdir $test_folder
+
+  # global.json file
+  cat >$test_folder/global.json <<EOF
+{
+  "sdk": {
+    "version": "$sdk"
+  }
+}
+EOF
+
+  # csproj file
+  cat >$test_folder/testapp.csproj <<EOF
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="$package" />
+  </ItemGroup>
+</Project>
+EOF
+
+  # Program.cs file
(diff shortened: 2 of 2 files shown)


tmds:
This tests https://bugzilla.redhat.com/show_bug.cgi?id=1588077

CC @omajid @RheaAyase

RheaAyase:
cool o.o
I was about to start writing something for this bz this week...

tool-dev-certs/test.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tool-dev-certs",
-  "version": "2.1.300",
+  "version": "2.1.301",
```
tmds: @RheaAyase did you force push this to master?

RheaAyase: Yes you can ignore that, I'll deal with it.

aspnetpatch-21/test.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "aspnetpatch-21",
+  "version": "2.1",
+  "versionSpecific": true,
```
RheaAyase: Is this for all of 2.1 or just 2.1.400? You can also use feature numbers (since like yesterday)
And do we want it tested only and only on this version, or any other future versions as well? (specific)

tmds: The test verifies things work on all 2.1 sdks.
We'll want to test a similar thing on 2.2 sdks. For now, the test is 2.1 specific.

RheaAyase: Okay, that gives me an idea - would having a $1 be the version, i.e. 2.1 help parametrize this test?

tmds: > would having a $1 be the version, i.e. 2.1 help parametrize this test?

Yes. It would be more informative to pass it as environment variables, e.g. DOTNET_TEST_VERSION is the runtime version.

Let's scope this PR to testing it on 2.1. You can explore how it can be made more generic for 2.2.

RheaAyase: Yup

aspnetpatch-21/test.sh
```diff
@@ -0,0 +1,113 @@
...
+
+    # create project folder
+    create_project_for_sdk_and_package $sdk $package
+
+    # publish application
+    cd $test_folder
+    dotnet publish -o out >/dev/null
+
+    # read the package version
+    package_version=$(cat out/*.deps.json | jq -r '.targets[".NETCoreApp,Version=v2.1"]["testapp/1.0.0"].dependencies["'$package'"]')
```
RheaAyase: > aspnetpatch-21:  /mnt/ramdisk/dotnet-bunny/dotnet-regular-tests/aspnetpatch-21/test.sh: line 63: jq: command not found
> aspnetpatch-21:  cat: write error: Broken pipe

@tmds I don't think there is `jq` installed in the VM, do you want me to make it happen or rather replace it with something else?

tmds: I find `jq` the best option for working with json in bash. Yes, please install it in the VM.

RheaAyase: Okay will be probably useful in the future for other stuff.

judovana: not much better then jq. Is it our share VM?

PR #20, opened 2018-07-07
Make aspnetpatch test more verbose

omajid:
It looks like that operations like `dotnet publish` can fail. Lets capture the output so we can see what's going on.

cc @tmds @RheaAyase

### DIFF
--- a/aspnetpatch-21/test.sh
+++ b/aspnetpatch-21/test.sh
@@ -58,7 +58,7 @@ for sdk in "$sdks"; do
 
     # publish application
     cd $test_folder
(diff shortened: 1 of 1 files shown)


tmds:
I'm curious why this would fail, does it show now?

RheaAyase:
@tmds replied to Omair in an email, apparently restore failed to find AspNetCore.App >= 2.1.2

tmds:
Ah, it fails for 2.1.2 because it isn't on nuget.org yet. I hadn't considered that.

PR #23, opened 2018-08-23
aspnetpatch-21: On RHEL, all sdks should use the latest ASP.NET Core version.

tmds:
ref https://bugzilla.redhat.com/show_bug.cgi?id=1588077

CC @omajid @RheaAyase

### DIFF
--- a/aspnetpatch-21/test.sh
+++ b/aspnetpatch-21/test.sh
@@ -3,8 +3,34 @@
 set -e
 set -x
 
-latest_aspnet_package=2.1.3
-sdks=$(dotnet --list-sdks | awk '{print $1}' | sort -ur | head -1)
+# Ensure we have an up-to-date value for latest_aspnet_package.
+dotnet_version=$(dotnet --version)
+dotnet_version_patch=${dotnet_version##*.}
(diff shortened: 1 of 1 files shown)


omajid:
The way we have packaged older versions, we currently have no way of going back and updating older SDK versions.

Also, source-build, when they stop producing a newer sdk also implicitly stop updating it to point to the latest asp.net core patch version.

tmds:
> The way we have packaged older versions, we currently have no way of going back and updating older SDK versions.

We don't need to the older packages. Setting `LatestPatchVersionForAspNetCoreApp2_1` and `LatestPatchVersionForAspNetCoreAll2_1` is picked up by the older packages and they will use those versions instead. That is what was proposed in https://bugzilla.redhat.com/show_bug.cgi?id=1588077 and this test was written to verify that.

Our s2i image defaults to the lowest sdk. If we are not setting these variables, it means it will pick an out-dated ASP.NET Core version.

So we either need to:
- set these variables
or
- change s2i to default to the latest

omajid:
Ah. I like option 1. But option 2 is something we should consider for future releases too.

PR #54, opened 2019-04-25
Blacklist RHEL from lttng test

omajid:
The test relies on tooling not available on RHEL out of box.

We should find out if it's possible to install lttng-sessiond on RHEL
in a maintainable way and then re-enable this test when we have that.

cc @tmds @RadoslavCap @RheaAyase

### DIFF
--- a/lttng/test.json
+++ b/lttng/test.json
@@ -6,5 +6,6 @@
   "type": "bash",
   "cleanup": true,
   "platformBlacklist":[
+    "rhel"
   ]
 }


RheaAyase:
> in a maintainable way

As in on our server / tests, or in rhel proper?

omajid:
In our CI/test servers. @tmds had pointers to unofficial RPMs that we could install to see if they helped.

Adding something to RHEL itself is a much larger commitment and requires several stakeholders to sign off on that. I definitely dont want to look into that for something like `lttng-tools` just for the sake of tests.

PR #59, opened 2019-06-05
Fix bash-completion test failing in a corner case

omajid:
If bash-completion test is the first test run and it's running on a
system where no cli commands have ever run, the first invocation of an
SDK command (which might be completion via dotnet suggest) will cause
suprious output. That spurious output fails completion.

So check results for completion after triggering (and igoring) a SDK
command.


RheaAyase:
cool o_o

PR #61, opened 2019-06-18
Add a test to make sure lldb/sos is functional

omajid:
Please note that some of the tests are disabled because they fail on the builds I am running it against and I dont want to block future builds where things don't get any worse.

### DIFF
--- a/debugging-sos-lldb-via-core/test.sh
+++ b/debugging-sos-lldb-via-core/test.sh
@@ -1,0 +1,233 @@
+#!/bin/bash
+
+# Check whether coredumps produced by .NET Core can be used by sos
+# successfully. This test uses the built-in CoreCLR sos support, not
+# the new `dotnet sos` global tool.
+
+set -euo pipefail
+IFS=$'\n\t'
+
+lldb-core () {
+    commands=()
+    while [[ "$#" -ne 0 ]]; do
+        command=$1
+        commands+=('--one-line' "${command}")
+        shift
+    done
+    lldb --batch \
+         --no-lldbinit \
+         -c "${coredump}" \
+         --one-line "plugin load ${framework_dir}/libsosplugin.so" \
+         "${commands[@]}"
+}
+
+version=$1
+
+set -x
+
+if ! command -v lldb ; then
+    echo "lldb is not installed"
+    exit 1
+fi
+
+# Create a dump
+
+rm -rf TestDir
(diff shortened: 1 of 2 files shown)


omajid:
cc @davemulford @RheaAyase

debugging-sos-lldb-via-core/test.sh
```diff
@@ -0,0 +1,233 @@
...
+lldb-core 'clrthreads' > lldb.out
+cat lldb.out
+thread_id=$(grep -A5 'ID OSID ThreadOBJ' lldb.out | tail -4 | grep -v 'Finalizer' | head -1 | awk '{print $1}')
+lldb-core "thread select ${thread_id}" 'clrstack' > lldb.out
+cat lldb.out
+ip=$(grep 'OS Thread Id:' lldb.out -A5 | tail -n3 | grep -v -F '[Prestub' | head -1 | cut -d' ' -f2)
+lldb-core "ip2md ${ip}" > lldb.out
+cat lldb.out
+if ! grep 'IsJitted' lldb.out ; then
+    echo 'sJitted field not found in ip2md'
```
aslicerh: nit - ctypo: IsJitted?

debugging-sos-lldb-via-core/test.sh
```diff
@@ -0,0 +1,233 @@
...
+lldb-core 'clrstack' > lldb.out
+cat lldb.out
+stack_pointer=$(grep 'OS Thread Id' lldb.out -A3 | tail -1 | awk '{ print $1}')
+lldb-core "dumpstack ${stack_pointer}"
+cat lldb.out
+
+echo "[eestack]"
+lldb-core 'eestack' > lldb.out
+cat lldb.out
+grep 'Child-SP         RetAddr          Caller, Callee' lldb.out
```
aslicerh: nit: Is the whitespace here going to stay consistent? Should something like \s\+ be used?

debugging-sos-lldb-via-core/test.sh
omajid: Thanks! Fixed.

debugging-sos-lldb-via-core/test.sh
omajid: Thanks! Fixed.

RheaAyase:
Thanks y'all! I guess that we can merge and run this thing.

PR #68, opened 2019-07-12
Fix race between dotnet run and pgrep in debugging-sos-lldb-via-core

omajid:
There was a race in the test case. We run `dotnet new`, `dotnet run &`
and then pgrep for `dotnet exec`. If the system running the test is
under load, then the implicit `dotnet build` might still be running by
the time we do a pgrep. At that point a `dotnet exec` process does not
exist. That puts the test in a funky state - the test has already failed
(because of set -e) but a background process is still running.

In CI systems this makes the test hang forever.

Fix that by:

- Reducing the amount of work that needs to be done in `dotnet run`

- Ignore pgrep failure and retry until we see a `dotnet exec` process

### DIFF
--- a/debugging-sos-lldb-via-core/test.sh
+++ b/debugging-sos-lldb-via-core/test.sh
@@ -38,11 +38,15 @@ cd TestDir
 
 dotnet new web
 sed -i -e 's|.UseStartup|.UseUrls("http://localhost:5000").UseStartup|' Program.cs
-dotnet run &
+dotnet build
+dotnet run --no-restore --no-build &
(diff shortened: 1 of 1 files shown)


judovana:
@RheaAyase  was that the issue we were observing fri/thu  ?

RheaAyase:
Yes, but there was a whole another issue with the jenkins nodes having slightly invalid configuration which made it quite difficult to see anything in the end :P

judovana:
Interesting! What was that configuration error?

PR #78, opened 2019-08-02
Add CI

omajid:
This adds a CI runner based on Azure Dev Ops

I simplified various tests to not create grandchildren where possible. It looks like docker hangs if grandchildren are created (and killed?).

Some gotchas:

1. `createdump` doesn't work in this restricted environment so all tests that use createdump (1 for .NET Core 2.1 and 2 for .NET Core 2.2) fail

2. Nodejs tests will break on Fedora rawhide. I removed them.

    template-test:  EXEC : gyp ERR! stack error : Python executable "/usr/bin/python" is v3.7.4, which is not supported by gyp. [/tmp/tmp.QlFfqcWExx/angular-template/angular-template.csproj]

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,5 +12,5 @@ git clone git@github.com:redhat-developer/dotnet-bunny.git && cd dotnet-bunny &&
 
 ### Dependencies
 
-Dependencies: jq bash-completion /usr/bin/readelf npm strace /usr/bin/free /usr/bin/lldb findutils which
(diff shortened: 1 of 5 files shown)


RadoslavCap:
Why we have choosen Azure DevOps?

omajid:
I wanted public CI that anyone could access. I don't know of too many options there so I went with what other .NET Core projects were using too. I am happy to switch to some other CI system if that works better for us.

RadoslavCap:
@omajid thank you for an explanation. I was just curious.

PR #79, opened 2019-08-23
Initial support for .NET Core 3.0

omajid:
With this set of patches, and running against .NET Core source-build on Fedora 30, I am down to:

`Total: 23 Passed: 21 Failed: 2`

The two failing tests are:

- `commit-ids-in-binaries`. This is a bug in .NET Core: https://github.com/dotnet/source-build/issues/1140
- `debugging-sos-lldb-via-core`. The debugger is failing to do _anything_. Needs further investigation.


RheaAyase:
Seems good thank you!

PR #84, opened 2019-09-03
Add tests for some Japanese Eras

omajid:
This was specially called out in the .NET Core 3.0 Preview 5 release
notes under "New Japanese Era (Reiwa)":
https://devblogs.microsoft.com/dotnet/announcing-net-core-3-0-preview-5/

I am not sure if it's supported on older .NET Core versions; it should be.

For more about calendars, see:

- https://devblogs.microsoft.com/dotnet/handling-a-new-era-in-the-japanese-calendar-in-net/
- https://docs.microsoft.com/en-us/dotnet/standard/datetime/working-with-calendars

### DIFF
--- a/japanese-eras/JapaneseEras.cs
+++ b/japanese-eras/JapaneseEras.cs
@@ -1,0 +1,39 @@
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using Xunit;
+
+namespace JapaneseEras
+{
+    public class JapaneseEras
+    {
+        [Theory]
+        [InlineData("2019-01-01", 4)] // Heisei
+        [InlineData("2019-04-30", 4)] // Heisei
+        [InlineData("2019-05-01", 5)] // Reiwa
+        [InlineData("2019-12-31", 5)] // Reiwa
+        public void VerifyEraIds(string date, int expectedEra)
+        {
+            var calendar = new JapaneseCalendar();
+            var time = DateTime.Parse(date);
+            int era = calendar.GetEra(time);
+            Assert.Equal(expectedEra, era);
+        }
+
+        [Theory]
+        [InlineData("2019-01-01", "平成")] // Heisei
+        [InlineData("2019-04-30", "平成")] // Heisei
+        [InlineData("2019-05-01", "令和")] // Reiwa
+        [InlineData("2019-12-31", "令和")] // Reiwa
+        public void VerifyEraNames(string date, string expectedEra)
+        {
+            CultureInfo japaneseCulture = new CultureInfo("ja-JP");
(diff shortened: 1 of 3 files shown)


tmds:
@omajid I wonder why we are testing this.

omajid:
@tmds I am trying to work my way through adding tests for all new 3.0 features. This seemed like it was applicable to 2.x as well. It also required changes in a bunch of other libraries/runtimes (glibc, openjdk, etc) so I wanted to make sure .NET Core had this correct too.

tmds:
> I am trying to work my way through adding tests for all new 3.0 features

Is this a goal? I think it makes sense to focus these tests on what could fail on rhel/fedora specifically.

omajid:
Good point; I will try and prioritize as you suggest.

tmds:
It's good your validating new features on platforms we care about. My concern is mostly that by adding tests in this repo we're raising the bar with things we expect to work at a release.

omajid:
I see what you are saying and I think this test definitely falls in the set of issues you think we should test:

1. This feature touches native code and system libraries. The complete fix on *nix relies on system libraries (glibc, libicu) as well as .NET Core. This is actually something that could fail on Fedora and RHEL if all components are not fully updated and configured correctly. AFAIK both RHEL and .NET Core needed updates to handle this.

2. This is fairly low-level stuff in corefx. If it's not working, we should be notified immediately. It's a regression test for 2.x as well.

tmds:
There's really a lot that fits this description.

If this test fails, do we postpone the release, or report an issue upstream. I'd put this in the later bucket, but maybe you'd put it in the first.

I'm also assuming we're covering this by running corefx unit tests.

omajid:
> There's really a lot that fits this description.

Hm. Any thoughts on how we can further narrow down the scope?

> If this test fails, do we postpone the release, or report an issue upstream. I'd put this in the later bucket, but maybe you'd put it in the first.

Honestly, probably the second. But we would need to investigate where the error happened. It could be a fault in a system library (glibc? libicu?) or in .NET Core itself.

> I'm also assuming we're covering this by running corefx unit tests.

I don't think we are :( I would be happy to be proven wrong.

tmds:
> Hm. Any thoughts on how we can further narrow down the scope?

Keep our tests at the tooling level (except for regressions). So using cli, template code, sos plugin, ...

omajid:
I just found out that this test does fail on RHEL 8.0 (`libicu` is too old). I think a fix will be included in RHEL 8.1. Some versions of RHEL 7 might also fail this test.

tmds:
Yes, the blog post mentioned this requires a recent version of ICU.

PR #90, opened 2019-09-30
Fix Runtime ID computation on RHEL

omajid:
RHEL 7.7, for example has a `VERSION_ID` of `7.7`. Using the computed runtime id of `rhel.7.7-x64` doesn't work, and doesn't seem correct either.

Centralize the logic used to generate Runtime IDs and handle RHEL there, specially.

cc @tmds


runtime-id
```diff
@@ -0,0 +1,48 @@
...
+        --portable) portable_rid=1 ;;
+        *) usage; exit 1 ;;
+    esac
+done
+
+if [[ ${portable_rid} == 1 ]]; then
+    echo "linux-${arch}"
+else
+    case "${ID}" in
+        rhel) rid_version=${VERSION_ID%.*} ;;
```
tmds: Can you add a comment? like: strip RHEL minor.

omajid: Done.

omajid:
@tmds The other thing we could do is actually add RHEL 7 runtime ids here:

https://github.com/dotnet/corefx/blob/9677a1458017796e4f42ae04c73cffce9e99aef2/pkg/Microsoft.NETCore.Platforms/runtimeGroups.props#L115

Does that seem like something worth doing?

tmds:
> Does that seem like something worth doing?

We should try to trim away the minors as close as possible to /etc/os-release reads. That stops them from propagating. Then we don't have to update the rid graph and other things.

PR #93, opened 2019-10-10
Add a test to check file permissions

omajid:
This should help us catch issues like apphost missing execute permissions for non-root users.

This currently fails for my 3.0 builds.

cc @tmds


RheaAyase:
Awesome :]

PR #95, opened 2019-11-01
Verify the certificate store path is valid

omajid:
See https://bugzilla.redhat.com/show_bug.cgi?id=1768029 for the bug we want to identify in the future.


omajid:
@tmds @RheaAyase, okay to merge?

Issue #96, opened 2019-11-13
Enable sos/lldb test on RHEL 8


omajid:
Opening an issue to make sure we don't forget about this: https://github.com/redhat-developer/dotnet-regular-tests/commit/3ab174bd8c202d8d13f9a6d305fd93ac58f29f35

omajid:
cc @tmds

tmds:
This doesn't crash for me on RHEL8 using .NET Core 3.0.

I had to remove this line:
https://github.com/redhat-developer/dotnet-regular-tests/blob/90d15f730671cf642945d413bb5be87f8020a794/debugging-sos-lldb-via-core/test.sh#L182
Seems like it is no longer part of the output.

I tried running against 2.1, but then I get a message 'not running compatible .NET Core runtime'. So the 'dotnet dump' tool doesn't support 2.1.
I'll re-enable the test for rhel8.

PR #97, opened 2019-11-13
Re-enable debugging-sos-lldb-via-core on rhel8

tmds:
Fixes https://github.com/redhat-developer/dotnet-regular-tests/issues/96

### DIFF
--- a/debugging-sos-lldb-via-core/test.sh
+++ b/debugging-sos-lldb-via-core/test.sh
@@ -32,8 +32,10 @@ sdk_version=$1
 
 set -x
 
-dotnet tool install -g dotnet-sos  --version 3.0.0-preview8 || true
-dotnet tool install -g dotnet-dump  --version 3.0.0-preview8 || true
+dotnet tool uninstall -g dotnet-sos || true
+dotnet tool uninstall -g dotnet-dump || true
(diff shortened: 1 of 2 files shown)


debugging-sos-lldb-via-core/test.sh
```diff
@@ -32,8 +32,10 @@ sdk_version=$1
 
 set -x
 
-dotnet tool install -g dotnet-sos  --version 3.0.0-preview8 || true
-dotnet tool install -g dotnet-dump  --version 3.0.0-preview8 || true
+dotnet tool uninstall -g dotnet-sos || true
+dotnet tool uninstall -g dotnet-dump || true
+dotnet tool install -g dotnet-sos  --version 3.0.0
+dotnet tool install -g dotnet-dump --version 3.0.0
```
omajid: I wonder what happens when 3.1 is released. Anyway, that's a question for another time.

tmds: Maybe we shouldn't specify a version here, to check whether the latest version of the tool works?

omajid: That sounds like a great idea!

PR #100, opened 2020-01-14
Add tests to make sure binaries are hardened

omajid:
As a bonus, this allows us to verify that the binaries were compiled by us and are not prebuilts accidentally leaked into our builds.

### DIFF
--- a/hardened-binaries/test.json
+++ b/hardened-binaries/test.json
@@ -1,0 +1,11 @@
+{
+  "name": "hardened-binaries",
+  "enabled": true,
+  "version": "3.0",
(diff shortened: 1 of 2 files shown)


omajid:
This change uses `readelf -l` and `readelf -d` to manually identify things.

Alternatively, we could use `checksec` as suggested here: https://www.redhat.com/en/blog/hardening-elf-binaries-using-relocation-read-only-relro. However, `checksec` is a separate package. But it is only a single stand-alone shell script so we can even download it a runtime, or even embed a copy if we need to.

I used `readelf` because it's more straightforward for now.

PR #101, opened 2020-01-14
Verify the host from the SDK is usable

omajid:
This is the other side of
https://github.com/dotnet/source-build/issues/1202. In this case the
apphost is not known and the SDK downloads the host from nuget.org
instead of using the host that is included with the sdk.

cc @tmds

### DIFF
--- a/use-apphost-from-sdk/test.json
+++ b/use-apphost-from-sdk/test.json
@@ -1,0 +1,11 @@
+{
+  "name": "use-apphost-from-sdk",
+  "enabled": true,
+  "version": "3.0",
+  "versionSpecific": false,
+  "type": "bash",
+  "cleanup": true,
+  "platformBlacklist":[
+
+  ]
+}
--- a/use-apphost-from-sdk/test.sh
+++ b/use-apphost-from-sdk/test.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+rm -rf console
+mkdir console
+pushd console
+
+# Verify that the apphost from the SDK is usable and nuget is not
+# needed for building a HelloWorld console application.
+
+cat > nuget.config <<EOF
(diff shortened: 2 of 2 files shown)


use-host-from-sdk/test.sh
```diff
@@ -0,0 +1,25 @@
...
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+  </packageSources>
+</configuration>
+EOF
+
+dotnet new console
+dotnet run --no-restore
```
tmds: nit: it may not be obvious for someone reading this test where/how this is using the sdk host.

omajid: Any suggestions on how to make this more obvoius? Do you think the test should explicitly tests for the host file here? Or is it more about explicitly documenting something?

omajid: Any thoughts, @tmds ?

tmds: You could explicitly use the apphost to launch the app:
```
dotnet build --no-restore
bin/Debug/netcoreapp*/console
```

use-host-from-sdk/test.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "use-host-from-sdk",
```
tmds: nit: `host` -> `apphost`

use-host-from-sdk/test.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
+rm -rf TestDir
+mkdir TestDir
+pushd TestDir
+
+# Verify that the host from the SDK is usable and nuget is not needed
```
tmds: nit: `host` -> `apphost`

omajid:
Does the new patch look okay?

Issue #103, opened 2020-01-20
Test shipped assemblies are compiled ahead-of-time


tmds:
With 3.0 and 3.1 servicing some assemblies are no longer being compiled ahead-of-time (https://bugzilla.redhat.com/show_bug.cgi?id=1791462, https://github.com/dotnet/source-build/issues/1452).

We should add a test to detect regressions.

We already have a test that looks into the assemblies: https://github.com/redhat-developer/dotnet-regular-tests/blob/master/dlls-in-release-mode/check-ngen-dlls. This test is written in Python. We may want to build further on that.

Alternatively, we can check using C#:
```cs
using System;
using System.IO;
using System.Reflection.PortableExecutable;

namespace console
{
    class Program
    {
        const int LINUX_IMAGE_FILE_MACHINE_NATIVE_OS_OVERRIDE = 0x7B79;

        static void Main(string[] args)
        {
            string assembly = args[0];

            PEReader reader = new PEReader(File.OpenRead(assembly));
            var managedNativeDirectoryHeader = reader.PEHeaders.CorHeader.ManagedNativeHeaderDirectory;
            bool r2r = managedNativeDirectoryHeader.Size != 0;
            Console.WriteLine($"R2R: {r2r}");
            if (r2r)
            {
                Machine machine = reader.PEHeaders.CoffHeader.Machine;
                Machine linuxArch = (Machine)((int)machine ^ LINUX_IMAGE_FILE_MACHINE_NATIVE_OS_OVERRIDE);
                Console.WriteLine($"Linux arch: {linuxArch}");
            }
        }
    }
}
```

Or maybe we want to combine both tests, and write them in C#?

I don't speak Python.

cc @omajid @RheaAyase

omajid:
I think we need to combine both tests into a single test that checks all our assemblies are valid. Written in C# itself.

Any thoughts or preferences @aslicerh and @RheaAyase ?

omajid:
A couple of questions:

- At least some `.dll` files (eg `System.Private.CoreLib.dll` in 2.1) have a `ManagedNativeHeader` that points to a `NGEN` header (that's what `check-ngen-dlls` looks at) instead of `R2R` header. Is it safe to ignore those?

- Which assemblies are supposed to be ready-to-run compiled?
  - All `.dll` files, even `.resource.dll` files?
  - Targeting packs?
  - Reference packages?
  - SDK tools?
  - Shared framework?
    - `mscorlib.dll`?
[...]

PR #104, opened 2020-02-05
Verify assemblies are built in READYTORUN mode

omajid:
Verify that assemblies are compiled in READYTORUN mode to catch issues like these: https://github.com/dotnet/source-build/issues/1452

This change merges the test previously written in python ("dlls-in-release-mode") with this new test to avoid duplication.

We are mostly interested in assemblies in the shared framework, not in the SDK or targeting packs.

Fixes: #103

cc @RheaAyase @tmds


RheaAyase:
Oh I missed this sorry, seems good to me. @tmds anything to add? :)

assemblies-valid/AssembliesValid.cs
```diff
@@ -0,0 +1,219 @@
...
+                p.WaitForExit();
+                using (StreamReader r = p.StandardOutput)
+                {
+                    standardOutput = r.ReadToEnd();
+                }
+                return p.ExitCode;
+            }
+        }
+
+        static Machine GetCurrentMachine()
```
tmds: At some point we may want to put these functions in a shared source folder, so multiple tests can use them.

assemblies-valid/AssembliesValid.cs
```diff
@@ -0,0 +1,219 @@
...
+        };
+
+        [Fact]
+        public void ValidateAssemblies()
+        {
+            string dotnetPath = null;
+            int exitCode = RunProcessAndGetOutput(new string[] { "command", "-v", "dotnet" }, out dotnetPath);
+            if (exitCode != 0)
+            {
+                Console.Error.WriteLine("'dotnet' command not found");
```
tmds: This test is a Console application in disguise :smile: 
General practice is not to do use `Console` in tests, and split them into different `[Fact]`/`[Theory]`. The `[Fact]` that failed, and the exception thrown should be enough to figure out what is failing.

I'm fine if you want to keep the test as is.

tmds:
Did you verify the test is failing as expected on an non-r2r build (ours), and passing on an r2r build (like Microsoft's)?

omajid:
> Did you verify the test is failing as expected on an non-r2r build (ours), and passing on an r2r build (like Microsoft's)?

Yes. For source-build, it flags all of 2.1.x as okay. It flags the 3.1.x `Microsoft.NETCore.App` as okay, and flags `Microsoft.AspNetCore.App` as being buggy. It flags all of  Microsoft's 3.1.x as okay.

omajid:
@tmds Any other concerns? Is this okay to merge?

tmds:
@omajid this is good to merge!

PR #110, opened 2020-06-05
Add requiresSdk json property

RheaAyase:


dotnet-info-commit-ids/test.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "dotnet-info-commit-ids",
   "enabled": true,
+  "requiresSdk": false,
```
omajid, line 4: The SDK includes some commit ids that we should check for:

```
$ dotnet --info
.NET Core SDK (reflecting any global.json):
 Version:   3.1.103
 Commit:    6f74c4a1dd

Runtime Environment:
 OS Name:     fedora
 OS Version:  32
 OS Platform: Linux
 RID:         fedora.32-x64
 Base Path:   /usr/lib64/dotnet/sdk/3.1.103/

Host (useful for support):
  Version: 3.1.3
  Commit:  4a9f85e9f8

.NET Core SDKs installed:
  3.1.103 [/usr/lib64/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.App 3.1.3 [/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.3 [/usr/lib64/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET Core runtimes or SDKs:
  https://aka.ms/dotnet-download
```

This test doesn't require an SDK, but the output of the test is less useful without the SDK installed.

file-permissions/test.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "file-permissions",
   "enabled": true,
+  "requiresSdk": false,
```
omajid, line 4: Same as above.

managed-symbols-available/test.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "managed-symbols-available",
   "enabled": true,
+  "requiresSdk": false,
```
omajid, line 4: I don't know if this is meaningful to run without the SDK. The symbols are part of the SDK.

PR #112, opened 2020-06-29
Disable system-libunwind on fedora-arm64

omajid:
The libunwind package on Fedora is broken for .NET Core to use on arm64. .NET Core uses the bundled copy, at least for now.


omajid:
This relies on a change in the test runner to work: https://github.com/redhat-developer/dotnet-bunny/pull/31

PR #117, opened 2020-07-10
Initial support for .NET 5

omajid:
This contains a grab bag of fixes:

- The framework has changed from `netcoreapp*` to `net5.0`. We need to shorten the string used to search for the framework in various tests.

- The Hello-World F# example uses different casing and has no exclamation mark at the end. Relax the string match requirements to support that.

- There are now multiple implementations of BadHttpRequestException available. Fully qualify it.

- Relax the version checks to support the bump from 3.x to 5.0.

With these fixes, there are still about 5 test failures. Some are bugs in source-build. Others would be better addressed via https://github.com/redhat-developer/dotnet-bunny/issues/34

### DIFF
--- a/apphost-framework-lookup/test.sh
+++ b/apphost-framework-lookup/test.sh
@@ -18,7 +18,7 @@ test_publish()
   mkdir $name && pushd $name
   dotnet new console
   dotnet publish -r "$runtime_id" --self-contained $2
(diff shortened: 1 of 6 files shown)


RheaAyase:
(Not merging until current release is out)

PR #118, opened 2020-07-16
Add logging to aspnet runtime version 2x test

omajid:
Add some logging information so we can see what's going on in case this test fails.

### DIFF
--- a/aspnet-same-runtime-version-2x/test.sh
+++ b/aspnet-same-runtime-version-2x/test.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+set -x
+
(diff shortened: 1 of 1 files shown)


omajid:
I spend a little bit more time looking into this. This is a bug in the test runner (`turkey`). When a bash tests fails very quickly, it doesn't capture the standard output or standard error from the test. 

I suspect there's a race or bug in the async code at https://github.com/redhat-developer/dotnet-bunny/blob/master/Turkey/ProcessExtensions.cs. The bug can be seen in the sample test too: https://github.com/redhat-developer/dotnet-bunny/blob/master/Samples/FailingBashTest/test.sh

PR #119, opened 2020-08-07
Centralize how .NET directoires are looked up

omajid:
The new script `dotnet-directory` allows everyone to consistently find the location of things related to .NET:

- The main installation directory

  Eg: /usr/lib64/dotnet

- The shared framework directory

  Eg: /usr/lib64/dotnet/shared/Microsoft.NETCore.App/3.1.5/

This makes it easier to modify things if the directory paths ever change. We should also get more consistent behaviour if more than one runtime version is installed.

### DIFF
--- a/bundled-libunwind/test.sh
+++ b/bundled-libunwind/test.sh
@@ -4,8 +4,10 @@ set -euo pipefail
 
 set -x
 
+framework_dir=$(../dotnet-directory --framework "$1")
+
 set +e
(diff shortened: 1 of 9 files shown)


RheaAyase:
When do you want to merge this? ASAP or after some things are dealt with first?

omajid:
No rush. I can wait a few days, or even a week or two for this.

PR #121, opened 2020-09-13
Update Microsoft.NET.Test.SDK for .NET 5 compatibility

omajid:
Preview 8 needs a newer version of Microosft.NET.Test.SDK reference package for tests to work at all. This is a known issue in Preview 8. See https://github.com/dotnet/core/blob/master/release-notes/5.0/5.0-known-issues.md#preview-8-1

It doesn't seem like a bad idea to just update everything to a newer version.

### DIFF
--- a/assemblies-valid/assemblies-valid.csproj
+++ b/assemblies-valid/assemblies-valid.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


omajid:
That original URL doesn't contain the issue, because someone removed it from the known issues since it was fixed in RC1. The issue exists in Preview 8. Here's a better link: https://github.com/dotnet/core/commit/fdfa548e39d5899ba1cbdb792b16cb40dacc6f47#diff-548c53695d8584aed2e2c83833e85d99

PR #123, opened 2020-09-14
extract-bundle-dir is only valid for 3.1

omajid:
This feature was introduced with .NET Core 3.0 (now EOL). In .NET 5.0, the single standalone file is not extracted to a on-disk location. This test is not needed there. Because the bundle extraction directory is empty, the test also results in a false positive.

See https://github.com/dotnet/runtime/issues/36590 for more information.

### DIFF
--- a/extract-bundle-dir/test.json
+++ b/extract-bundle-dir/test.json
@@ -3,7 +3,7 @@
   "enabled": true,
   "requiresSdk": true,
   "version": "3.1",
-  "versionSpecific": false,
+  "versionSpecific": true,
   "type": "bash",
   "cleanup": true,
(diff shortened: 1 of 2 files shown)


omajid:
@tmds What do you think? Does this change make sense? Is there anything I missed about bundle extraction?

omajid:
Hey, @tmds, any objections to merging this? As I understand it, the bundle-extraction only happens on 3.1 so the test not useful for 5.0 and later.

tmds:
> the bundle-extraction only happens on 3.1 so the test not useful for 5.0 and later.

Yes, we only need to run this test for 3.1.

PR #125, opened 2020-09-18
Handle some packages having different major/minor versions

omajid:
This test checks all packages have the same exact version. However, when there are multiple packages available, dotnet-host is provided from the latest .NET version while dotnet-runtime-X.Y are version specific.

Handle dotnet-host and netsandard-targeting-pack-X.Y having different verisons.


omajid:
fyi @RheaAyase

PR #126, opened 2020-09-19
Update versions in CI

omajid:
Update Fedora versions to 32, 33 and Rawhide. Fedora 30 is already EOL and Fedora 31 will be soon. Fedora 32 also includes .NET Core 3.1 in its default repositories.

.NET Core 2.2 and 3.0 are also EOL. Only 2.1 and 3.1 are fully supported right now.

Also save and publish test run results.

Everything should now be passing: https://dev.azure.com/omajid/omajid/_build/results?buildId=167&view=results

Test result example: https://dev.azure.com/omajid/omajid/_build/results?buildId=167&view=ms.vss-test-web.build-test-results-tab

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,5 +12,5 @@ $ git clone https://github.com/redhat-developer/dotnet-bunny && cd dotnet-bunny 
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


RheaAyase:
Oh... looks good :D

PR #129, opened 2020-10-07
Define used package versions as properties

tmds:
This consolidates versions of packages used in `Directory.Build.props`.

It also removes some unused files from `libuv-kestrel-sample-app-2x`.

It also removes the `libuv-kestrel-sample-app` test. This 3.x test has an explicit version for a Libuv package. It is not a common case to use this package anymore, and we don't build and ship it ourselves. I think we can remove it for those reasons.


RheaAyase:
Cool :)

PR #130, opened 2020-10-08
Add xunit smoke test

tmds:


omajid (approved):
Oh, this is brilliant!

I was looking at how Preview 8 sometimes passes tests (even though no tests are found) and sometimes fails tests (if right libraries are not used). This way of verifying both passing and failing sounds great!

xunit-smoketest/test.sh
```diff
@@ -1,10 +1,12 @@
...
 
+set -euo pipefail
+
 for TEST_RESULT in pass fail ;
 do
   export TEST_RESULT
-  OUTPUT=$(dotnet test 2>&1)
-  EXIT_CODE=$?
+  EXIT_CODE=0
+  OUTPUT=$(dotnet test 2>&1) || EXIT_CODE=$?
```
omajid, line 9: Maybe a `;` should be better instead of `||`? We want to always set the exit code, don't we?

tmds: The ` || ` is to avoid exiting immediately when `dotnet test` fails (due to `set -e`).

PR #134, opened 2020-12-22
Add a test to verify the SDK/Runtime version is sane

omajid:
This makes sure the SDK/Runtime version is either the latest version, or 1 patch version newer than the latest version available publicly.

This should help catch any cases where we are accidentally releasing a too-new or too-old version.

### DIFF
--- a/release-version-sane/UpstreamRelease.cs
+++ b/release-version-sane/UpstreamRelease.cs
@@ -1,0 +1,76 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


RheaAyase:
+1

RheaAyase:
Hmmm we could use a feature where we mark tests as warning only on failure, not a complete failure. That would complicate a lot of things though :D

omajid:
I am going to merge it now, but with the expectation that something strange might pop up that would require us to revert this test, or filter it for some specific platfroms or use-cases.

PR #135, opened 2021-01-29
Extend omnisharp test to cover more project types

omajid:
Omnisharp needs separate tasks/dlls for separate project types. Testing with just a `console` project is not sufficient to verify that all project types work. We need to test with each project type that we care about to make sure it works.

See https://github.com/dotnet/source-build/issues/2006 for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
     timeout-minutes: 30
 
     strategy:
+      fail-fast: false
       matrix:
         container_image:
           - fedora:34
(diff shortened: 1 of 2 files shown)


omajid:
Not sure why CI is breaking, but this test does fail locally right now because of the underlying bug. We should wait until we have a fix before merging this test.

omajid:
https://github.com/dotnet/source-build/pull/2178 was merged, so I am going to merge this too to make sure future .NET 5 releases have a fix for this.

cc @RheaAyase

Issue #108, opened 2020-03-20
Test ProcessorCount reflects cpu container constraints


tmds:
.NET Core changed ProcessorCount in 3.1.0 to no longer reflect cpu container constraints.
The change was intentional, but lead to unexpected performance regressions in containers.
.NET Core 3.1.2 reverted back to the previous implementation.

cc @omajid @RheaAyase

PR #137, opened 2021-03-22
Add cgroup-limit test

tmds:
Fixes https://github.com/redhat-developer/dotnet-regular-tests/issues/108

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           set -euo pipefail
           ### HACK: Filter tests that can't pass in Containers
-          rm -rf debugging-sos-lldb* createdump-aspnet
+          rm -rf debugging-sos-lldb* createdump-aspnet cgroup-limit
 
           ### HACK: UBI 8 is missing strace and bash-completion packages for tests
           if [[ ${{ matrix.container_image }} == *ubi8* ]] ; then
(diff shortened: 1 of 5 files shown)


omajid:
CI is failing because this CI runs within a container and can't use `systemd-run`:

```
++ systemd-run -q --scope --user -p CPUQuota=100% -p MemoryLimit=100M bin/Debug/netcoreapp3.1/cgroup-limit
Failed to create bus connection: No such file or directory
```

For now, can you add a hack here? https://github.com/redhat-developer/dotnet-regular-tests/blob/bf7f0eabdb2f033f9cb9d87699c94b68f471fc77/.github/workflows/ci.yml#L54-L55

tmds:
@omajid I've excluded the test from the github ci. The test passes on my Fedora 33. Do we have some automation that will run this new test on RHEL7 and RHEL8?

omajid:
I don't think we have pre commit automation.

@RheaAyase do you have a way to run this pre commit? Or do we need to merge this before we can see how well it runs?

tmds:
@omajid @RheaAyase I want to be sure that something runs this test on RHEL7/RHEL8 before it gets the opportunity to fail when we make a release. I'm fine if it runs when this is merged. The test itself is fine, since it works for me on Fedora 33.

tmds:
From our irc discussion: CI will test this on RHEL8, and RHEL7. So I'll merge this.

PR #143, opened 2021-04-19
Show cgroup system information

omajid:
That will help us figure out what cgroup system the test runner is using. That's useful to know because .NET Core 3.1 doesn't fully support cgroupv2.


tmds:
> .NET Core 3.1 doesn't fully support cgroupv2.

Do we want to update the test to reflect this?

PR #145, opened 2021-04-19
Don't test cgroup support with .NET Core 3.1 on cgroup v2

omajid:
.NET Core 3.1 didn't support cgroup v2, so it's known not to work. It should still work with cgroup v1. Later versions of .NET (eg, .NET 5) should work in both cgroup v1 and v2.


omajid:
cc @tmds @RheaAyase

PR #146, opened 2021-04-29
Check offline framework-dependent deployment

HoundThe:
I've tried to add the test to check the FDD deployment. Currently, there is a test just for 5.0, I guess I need to make a separate test case for 3.1 as well and can't reuse it (from what I've seen in other tests in the repository)?

### DIFF
--- a/fdd-publish-5.0/fdd-publish.csproj
+++ b/fdd-publish-5.0/fdd-publish.csproj
@@ -1,0 +1,6 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
+  </PropertyGroup>
+</Project>
--- a/fdd-publish-5.0/test.json
+++ b/fdd-publish-5.0/test.json
@@ -1,0 +1,13 @@
+{
+	"name": "fdd-publish-5.0",
+	"enabled": true,
+	"requiresSdk": true,
+	"version": "5.0",
+	"versionSpecific": true,
+	"type": "bash",
+	"cleanup": true,
+	"ignoredRIDs":[
+  
+	]
+  }
+  
--- a/fdd-publish-5.0/test.sh
+++ b/fdd-publish-5.0/test.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/bash
+
+set +x
+
+dotnet publish
--- a/fdd-publish-5.0/Program.cs
+++ b/fdd-publish-5.0/Program.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace fdd_publish
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
+        }
+    }
+}
--- a/fdd-publish-5.0/nuget.config
+++ b/fdd-publish-5.0/nuget.config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+  </packageSources>
+</configuration>


fdd-publish-5.0/fdd-publish.csproj
```diff
@@ -0,0 +1,6 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
```
tmds, line 4: `netcoreapp5.0` -> `net5.0`

fdd-publish-5.0/test.json
```diff
@@ -0,0 +1,13 @@
+{
+	"name": "fdd-publish-5.0",
```
tmds, line 2: `fdd-publish-5.0` -> `fdd-no-nuget`

fdd-publish-5.0/test.json
```diff
@@ -0,0 +1,13 @@
...
+	"enabled": true,
+	"requiresSdk": true,
+	"version": "5.0",
```
tmds, line 5: `5.0` -> `2.1`

fdd-publish-5.0/test.json
```diff
@@ -0,0 +1,13 @@
...
+	"versionSpecific": true,
```
tmds, line 6: `true` -> `false`

fdd-publish-5.0/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/bash
+
+set +x
+
+dotnet publish
```
tmds, line 5: Can you make this print a PASS/FAIL message like the other tests?

fdd-publish-5.0/test.json
tmds: This will make the test run 2.1+.

fdd-publish-5.0/fdd-publish.csproj
tmds: Part of the test runner will automagically update the target framework in this file.

tmds:
@HoundThe thanks for picking this up, ptal at the comments.

HoundThe:
Thanks a lot for the review, I've noticed that when I run Turkey, it warns me that it overwrites the `nuget.config` file and it looks from the source code that Turkey appends another package sources for dotnet and deletes the config after. It looks like this after the modification from Turkey.
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
  </packageSources>
  <disabledPackageSources />
</configuration>
```
I guess moving the nuget.config creation to the shell file so it overwrites the Turkey one would be the solution?

HoundThe:
I've tried the `nuget.config` creation from a shell script and it seems to work.

fdd-no-nuget/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/bash
+
+set +x
+
```
tmds, line 4: Can you add a `rm -rf ~/.nuget`?

I'm not sure it is implied by any of the `test.json` properties.

fdd-no-nuget/test.sh
```diff
@@ -0,0 +1,22 @@
...
+  <packageSources>
+    <clear />
+  </packageSources>
+</configuration>
+EOF
+
+dotnet publish
+
+if [ $? -eq 1 ]; then
+  echo "FAIL"
```
tmds, line 18: "FAIL" -> "FAIL: unable to publish FFD without nuget.org access"
"PASS" -> "PASS: successfully published FDD without nuget.org access"

fdd-no-nuget/test.sh
omajid: `rm -rf ~/.nuget` is always implicit: https://github.com/redhat-developer/dotnet-bunny/blob/770c4a06f958072d935e7b88e77126a12fe469e8/Turkey/TestRunner.cs#L84

tmds: @HoundThe can you add it anyhow. It will no-op when used with the test runner. It's still useful when the test is ran by calling `./test.sh` directly.

tmds: I see you've added `rm nuget.config`, I meant: add a `rm -rf ~/.nuget` line.

tmds:
@HoundThe sorry for the slow feedback on the PR. This is good to merge when you've added the `rm -rf ~/.nuget`.

fdd-no-nuget/test.sh
HoundThe: Oh, lol. Sorry, fixed it.

PR #149, opened 2021-07-27
Add s390x arch to runtime-id

BahaVv:
This PR adds the s390x field to runtime-id to correct a failure within certain tests, including the apphost-framework-lookup test.


omajid:
Test failures are expected: rawhide is in a strange state with containers and stable releases of Fedora don't have the latest update yet.

PR #150, opened 2021-08-09
Test all templates in template-test

omajid:
The current test hardcodes a few templates and only tested those. New templates (eg, `dotnet new worker`) got ignored.

This change makes sure we run `dotnet new template` for all templates listed in the `dotnet new --list` output. For some specific templates, we will also run `dotnet build` or `dotnet run`, like before.

Also, explictly check that `dotnet new template` completes without restore errors. Restore errors, for some reason, dont cause `dotnet new` to exit with an error code. 

There are some alternative we could do too, if that sounds better:

- Keep a complete list of all templates that should be tested, but have that for each .NET version
  - Maybe complain if `dotnet new --list` shows a template not known?
- Other ideas?


omajid:
@RheaAyase WDYT?

RheaAyase:
Sounds cool!

omajid:
@RheaAyase any objections to merging this?

RheaAyase (approved):

PR #157, opened 2021-08-30
Add optional matches to helloworld for preview 7

BahaVv:
The hello world text for preview 7 has been adjusted from "Hello World!" to "Hello, World!" which breaks the helloworld-2x test. This adds optional characters through regex to allow the tests to be compatible with both variants.


omajid (approved):
LGTM!

This looks like the upstream change that changed things from `Hello World!` to `Hello, World!`: https://github.com/dotnet/templating/commit/73902e7909f949bc1ab10405328d0b18713b5cec#diff-79f14e93e40296be130cbc2a92583b0522f70abb362d6f9ca89068e68a17576c

PR #159, opened 2021-09-21
Disable createdump test on s390x

omajid:
s390x doesn't have a `createdump` utility.

```
dotnet-regular-tests/createdump-aspnet/test.sh: line 30: dotnet-sdk-s390x/shared/Microsoft.NETCore.App/6.0*/createdump: No such file or directory
```

So the test can not possibly pass. Disable the test on s390x.

cc @BahaVv

### DIFF
--- a/createdump-aspnet/test.json
+++ b/createdump-aspnet/test.json
@@ -7,5 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
   ]
 }


omajid:
Test failures are in Fedora, because the Fedora .NET packages are one-month-old. We can't update those to the September release until source-build does a tag/release. There aren't any other failures that I can see.

omajid:
Filed https://github.com/redhat-developer/dotnet-regular-tests/issues/164 to track re-enabling this test at some point in the future.

PR #161, opened 2021-09-27
More thorough man page test

omajid:
Show all man pages that are missing on first run. Also check that man pages for bundled tools exist.


omajid:
Uh.. CI says this test fails on all released versions and would prevent us from updating it. We shouldn't have merged this.

PR #165, opened 2021-09-27
Update version-apis test for .NET 6

omajid:
It only knows about .NET versions until 5, so tell it about 6 as well.


omajid:
Fedora errors are all in one (unrelated) test because the Fedora package version is a little old.

PR #167, opened 2021-09-29
man-pages should tell us which man pages were missing

omajid:
The test can currently fail silently. Also, show all the errors on the run instead of failing on the first error.

This is another attempt at #161, but hopefully that causes no regressions. Notably, compared to #161, we are *not* testing that man pages exist for bundled tools (`dotnet devcerts`, `dotnet fsi`, `dotnet user-secrets`, `dotnet watch`, and so on).


omajid:
CI test failures are all in `release-version-sane` which is failing because Fedora's packages are a few weeks old now.

PR #183, opened 2022-01-26
OmniSharp upstream broke compat with RHEL 7

omajid:
So use an older version of mono if the current wont work.


omajid:
cc @RheaAyase

PR #188, opened 2022-02-08
Fix/skip tests to improve test pass rate on s390x

omajid:
My goal is to get the default out-of-the-box state to fully pass tests
on s390x by default. Let's disable tests that are not supported on s390x
and fix the ones that should work to actually work correctly.

~~This is first of a series of fixes needed to make that happen.~~

    Total: 52 Passed: 33 Failed: 0

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <XunitVersion>2.4.1</XunitVersion>
     <XunitRunnerVisualStudioVersion>2.4.3</XunitRunnerVisualStudioVersion>
-    <MicrosoftNetTestSdkVersion>16.6.1</MicrosoftNetTestSdkVersion>
+    <MicrosoftNetTestSdkVersion>17.0.0</MicrosoftNetTestSdkVersion>
   </PropertyGroup>
 </Project>
--- a/apphost-framework-lookup/test.json
+++ b/apphost-framework-lookup/test.json
@@ -7,6 +7,7 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
   ]
 }
 
--- a/cgroup-limit/test.json
+++ b/cgroup-limit/test.json
@@ -7,5 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
   ]
 }
--- a/hardened-binaries/test.sh
+++ b/hardened-binaries/test.sh
@@ -8,7 +8,7 @@ echo ".NET Core base directory: $root"
 
 # TODO handle more architectures can just x86-64
 
-file_list=$(find "$root/" -type f -exec file {} \; | grep 'ELF 64-bit LSB' | cut -d: -f 1 | sort -u)
+file_list=$(find "$root/" -type f -exec file {} \; | grep -E 'ELF 64-bit [LM]SB' | cut -d: -f 1 | sort -u)
 mapfile -t binaries <<< "$file_list"
 for binary in "${binaries[@]}"; do
     echo "$binary"
--- a/liblttng-ust_sys-sdt.h/test.json
+++ b/liblttng-ust_sys-sdt.h/test.json
@@ -7,6 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-    "fedora31"
+    "linux-s390x"
   ]
 }
--- a/omnisharp/test.json
+++ b/omnisharp/test.json
@@ -6,6 +6,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-
+    "linux-s390x"
   ]
 }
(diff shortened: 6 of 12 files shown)


Directory.Build.props
```diff
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <XunitVersion>2.4.1</XunitVersion>
     <XunitRunnerVisualStudioVersion>2.4.3</XunitRunnerVisualStudioVersion>
-    <MicrosoftNetTestSdkVersion>16.6.1</MicrosoftNetTestSdkVersion>
+    <MicrosoftNetTestSdkVersion>17.0.0</MicrosoftNetTestSdkVersion>
```
omajid, line 5: This upgrades `Micrsoft.NET.Test.SDK` to a version that supports s390x.

apphost-framework-lookup/test.json
```diff
@@ -7,6 +7,7 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
```
omajid, line 10: This tests does a publish using `linux-s390x` on Linux. That's not supported, so disable this test.

cgroup-limit/test.json
```diff
@@ -7,5 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
```
omajid, line 10: Cgroup support is not implemented in mono, so skip this test on s390x.

hardened-binaries/test.sh
```diff
@@ -8,7 +8,7 @@ echo ".NET Core base directory: $root"
 
 # TODO handle more architectures can just x86-64
 
-file_list=$(find "$root/" -type f -exec file {} \; | grep 'ELF 64-bit LSB' | cut -d: -f 1 | sort -u)
+file_list=$(find "$root/" -type f -exec file {} \; | grep -E 'ELF 64-bit [LM]SB' | cut -d: -f 1 | sort -u)
```
omajid, line 11: `file` reports `LSB` on x86_64 and reports `MSB` on s390x (and other big-endian systems).

liblttng-ust_sys-sdt.h/test.json
```diff
@@ -7,6 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-    "fedora31"
+    "linux-s390x"
```
omajid, line 10: Mono doesn't use lttng, so this test is not useful at all.

omnisharp/test.json
```diff
@@ -6,6 +6,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-
+    "linux-s390x"
```
omajid, line 9: Omnisharp isn't available for s390x. Skip this test.

publish-ready-to-run/test.json
```diff
@@ -7,5 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
```
omajid, line 10: Publish ready-to-run isn't supported on s390x.

use-current-runtime/test.json
```diff
@@ -7,5 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
```
omajid, line 10: This test makes sure that a restore/build with `--use-current-runtime` uses the portable RID id, `linux-s390x`, which isn't going to work because those assets are not available publicly.

restore-with-rid/test.json
```diff
@@ -7,5 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
+    "linux-s390x"
```
omajid, line 10: This test does a:

     dotnet restore -r rhel.8-s390x

Which fails with:

```
restore-with-rid.csproj : error NU1101: Unable to find package Microsoft.NETCore.App.Runtime.linux-s390x. No packages exist with this id in source(s): [... ]
restore-with-rid.csproj : error NU1101: Unable to find package Microsoft.AspNetCore.App.Runtime.linux-s390x. No packages exist with this id in source(s): [...]
```

Is this expected?

omajid: Any thoughts @tmds ?

tmds: > Is this expected?

Yes, the sdk doesn't contain these nupkgs, and nuget.org doesn't have them either.

cgroup-limit/test.json
tmds: Maybe create an issue for re-enabling the test so we don't forget about it.

tmds (approved):

BahaVv:
Checked in a RHEL 8.4 environment using a 6.0.0 standalone SDK. 

The workload test fails with a permission denied error owing to wanting a very specific user named `testrunner`, which the system did not have.

Other than that, the only tests that failed in this environment are exactly what you would expect to - packaging tests.

So, LGTM.

PR #190, opened 2022-03-04
Guard use of Architecture.S390x on .NET 6 and later only

omajid:
It was only introduced in that version; trying to use it on .NET 5 or
earlier leads to build errors.

### DIFF
--- a/assemblies-valid/AssembliesValid.cs
+++ b/assemblies-valid/AssembliesValid.cs
@@ -90,11 +90,15 @@ namespace AssembliesValid
                         }
                         if (hasMethods && !hasAot)
                         {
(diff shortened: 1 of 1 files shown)


omajid:
Oops. The CI job is broken because we changed the release format in dotnet-bunny. Need to rebase on top of https://github.com/redhat-developer/dotnet-regular-tests/pull/191

omajid:
I am happy with CI now: this test is passing everywhere.

Any objections to merging this, @RheaAyase @BahaVv?

omajid:
Since this is time-sensitive, I am merging this. If you have any concerns, please file an issue or ping me and we can work on improving this test.

PR #196, opened 2022-04-20
Skip system-libunwind on Fedora s390x

omajid:
Like arm64, s390x uses a bundled libunwind.

cc @BahaVv


BahaVv:
This makes sense -- LGTM.

PR #199, opened 2022-06-23
Add a test to make sure console logging doesn't include colors

omajid:


omajid:
cc @tmds

tmds:
Upstream has a test for this also, so I don't think we need to check this.

The test looks good. Feel free to include it if you want.

omajid:
> Upstream has a test for this also, so I don't think we need to check this.

Huh. And they still managed to regress this feature? :laughing:

tmds:
The test got added with the fix.

PR #200, opened 2022-08-29
Extend OmniSharp tests to cover more versions

omajid:
OmniSharp upstream has been making a number of changes, and we should
extend this test to stay compatible with it. In particular, there a
number of different OmniSharp versions that were released recently and
all of them work differently in terms of dependencies they need and the
command line to use.

The latest releases (1.39 on) come in two flavours: A .NET 6 build and a
Mono-based-build. The Mono version requires an installation of mono on
the system to work. So download and run the .NET 6 version explicitly.

The slightly-older releases (1.38) include a bundled Mono. It needs to
use the ./run script to run. This mono is built against a recent version
of glibc that may not work well on some OS's.

The even-older releases (1.37) use a bundled Mono and need a script at a
different location to run.

Test all these variants of OmniSharp to ensure they all work, as much as
possible..

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,17 +18,22 @@ jobs:
       fail-fast: false
       matrix:
         container_image:
-          - registry.fedoraproject.org/fedora:34
           - registry.fedoraproject.org/fedora:35
+          - registry.fedoraproject.org/fedora:36
+          - registry.fedoraproject.org/fedora:37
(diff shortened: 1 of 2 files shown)


omajid:
Two failures in CI. One is because Fedora 37 is on an old version of .NET and the second I reported as https://github.com/redhat-developer/dotnet-regular-tests/issues/202

PR #207, opened 2022-09-16
Add smoke test for NativeAot.

tmds:
@omajid ptal.

### DIFF
--- a/nativeaot/test.sh
+++ b/nativeaot/test.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+set -x
+
+rm -rf bin obj
+
+# Publish the app.
+if ! dotnet publish --use-current-runtime /p:PublishAot=true; then
+    echo "FAIL: failed to publish application using NativeAot."
+    exit 1
+fi
+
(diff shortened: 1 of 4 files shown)


nativeaot/test.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+# Verify the published output contains a single file.
+PUBLISHED_FILE_COUNT=$(ls ./bin/Debug/net*/*/publish | wc -l)
+if [ "$PUBLISHED_FILE_COUNT" != "1" ]; then
+    echo "FAIL: published NativeAot app contains more than 1 file."
+    exit 1
+fi
+
+# Verify the application runs successfully.
```
omajid, line 22: Would it be useful to verify the application also produces the expected output and no other warnings/output?

omajid:
THANKS!!!!

nativeaot/test.sh
tmds: I've updated the test to verify the application output instead of checking the exit code.

tmds:
Is this good to merge @omajid ?

omajid:
Merged now!

Issue #208, opened 2022-09-19
Tracker for tests passing on .NET 7 


omajid:
Tests that need to be investigated and/or fixed before we feel comfortable releasing .NET 7

- [ ] `omnisharp`
- [ ] `telemetry-is-off-by-default`
- [ ] `template-test`
- [ ] `use-current-runtime`
- [ ] `version-apis`

PR #209, opened 2022-09-26
Get tests to mostly pass against .NET 7 RC 1

omajid:
- When parsing versions, split on both `.` and `-` to get all the version parts correctly, including any `-preview.` and `-rc.n` parts.

- In release-version-sane, ignore any prereleases such as -rc and -preview. Treat them as ordinary versions without the suffixes. So `7.0.0-rc.1.buildid` becomes `7.0.0`.

- Use `dotnet new web --no-restore` in the strace test to make sure build servers don't get started.

- Add templates for .NET 7.

- Allow 7.0 as a major .NET version in version-apis.

There's one known failures after these changes: `omnisharp`, which doesn't seem to support .NET 7.

See: #208

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,9 @@ jobs:
         dotnet_version:
           - "3.1"
           - "6.0"
+        include:
+          - container_image: quay.io/centos/centos:stream9-development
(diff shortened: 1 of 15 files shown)


omajid:
FYI, @janani66

PR #213, opened 2022-10-21
Added ppc64le in cases which needs to skipped and also in runtime-id

Swapnali911:
Added ppc64le architecture in ignoredRIDs for cases which requires nuget packages that are not shipped to the customer referring to s390x.

Also added in runtime-id, to run test on ppc64le

### DIFF
--- a/assemblies-valid/AssembliesValid.cs
+++ b/assemblies-valid/AssembliesValid.cs
@@ -90,15 +90,18 @@ namespace AssembliesValid
                         }
                         if (hasMethods && !hasAot)
                         {
+                            // 32-bit arm doesn't have AOT and niehter do s390x and ppc64le (which use mono). That's okay for now.
+                            if (architecture != Architecture.Arm
+#if NET7_0_OR_GREATER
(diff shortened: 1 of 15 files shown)


Swapnali911:
@omajid Kindly review.

assemblies-valid/AssembliesValid.cs
```diff
@@ -149,6 +149,8 @@ static Machine GetCurrentMachine(Architecture arch)
 #if NET6_0_OR_GREATER
                 case Architecture.S390x:
                     return Machine.Unknown;
+		case Architecture.Ppc64le;
```
omajid, line 152: Do we need to guard this with `#if NET7_0_OR_GREATER`? Otherwise .NET 6 will try and compile this code too. Is `Ppc64le` present in .NET 6?

janani66: Ppc64le is not present in .NET6 , so yet we need to guard this with NET7_0_OR_GREATER

janani66 (commented):
Reviewed the changes. Looks good.

PR #214, opened 2022-10-24
Update Microsoft.NET.Test.SDK for ppc64le support

omajid:
Only recent nightly builds of Microsoft.NET.Test.SDK support ppc64le. These very recent builds are only available via the development feeds (not on nuget.org), but the test runner already knows about them.


omajid:
Oh. It's only currently available in feeds for .NET 6 and later :/

PR #215, opened 2022-10-25
Hardcode a Microsoft.NET.Test.Sdk version with ppc64le enablement

janani66:
Fix the SDK version in template-test for ppc64le


omajid (approved):
Nice catch!

PR #216, opened 2022-11-01
Enable cgroup-limit test for s390x and ppc64le 

janani66:
The https://github.com/dotnet/runtime/pull/77682 is a prerequisite to enabling the cgroup-limit test

### DIFF
--- a/cgroup-limit/test.json
+++ b/cgroup-limit/test.json
@@ -7,7 +7,5 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-    "linux-s390x",
-    "linux-ppc64le"
   ]
 }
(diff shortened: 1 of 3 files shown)


cgroup-limit/test.json
```diff
@@ -7,7 +7,5 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-    "linux-s390x",
-    "linux-ppc64le"
```
omajid, line 11: I am not sure if this is the best way to go about this. This would enable the tests s390x and ppc64le even on older .NET releases, like 6.0. We should still skip the test for 6.0, somehow.

omajid:
Obsoleted by #224, #225 and #226

PR #224, opened 2022-12-14
Enable "cgroup-limit" dotnet-regular test on s390x on .NET 7.0 onwards

Vishwanatha-HD:
Enabling "cgroup-limit" dotnet-regular test on s390x on .NET 7.0 onwards. The support for cgroup-limit checking is not there in .NET 6.0 and earlier releases. So, this testcase will only be run from .NET 7.0 onwards.

PR #225, opened 2022-12-14
Add support for cgroup test on s390x

omajid:
Remove the s390x from unsupported architecture RID list

Enable "cgroup-limit" test on s390x on .NET 7.0 onwards

The feature support is provided from .NET 7.0 and beyond. This test should not run if the .NET release is 6.0 or earlier ones.

### DIFF
--- a/cgroup-limit/test.json
+++ b/cgroup-limit/test.json
@@ -7,7 +7,6 @@
   "type": "bash",
   "cleanup": true,
   "ignoredRIDs":[
-    "linux-s390x",
     "linux-ppc64le"
   ]
 }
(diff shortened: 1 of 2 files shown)


omajid:
@Vishwanatha-HD this is the same as #224. I accidentally cloned that one while trying to merge the 21 commits (about 19 of which were "Update test.sh") into 1.

Vishwanatha-HD:
Sorry for all the confusion.. I was trying to fiddle with syntax to get the exact command and thats how I ended up saving so many times.. I didnt realize that it would all become commits.. I tried removing them and retaining only the last and final ones.. But I was not successful.. Thats how I ended up raising a PR with 21 commits.. :-).. Sorry for the mess.. !!

Vishwanatha-HD:
@omajid  I saw the code merge that you have done finally.. Thanks for that.. !!

PR #226, opened 2022-12-14
Enable cgroup-test on ppc64le for .NET 7 and later

omajid:


omajid:
FYI @janani66 @Swapnali911 @Vishwanatha-HD

PR #222, opened 2022-12-07
Update distro matrix for CI

omajid:

PR #228, opened 2023-01-10
Add a test for System.Data.Odbc

omajid:
Using postgres


Swapnali911:
@omajid  I have validated for ppc64le RHEL8 and RHEL9
Testing everything under /home/testrunner/dotnet-regular-tests/system-data-odbc
Current platform is: linux, linux-ppc64le, rhel, rhel-ppc64le, rhel8, rhel.8, rhel.8-ppc64le
system-data-odbc                                            [**PASS**]      (12460.3796ms)

Testing everything under /home/dotnet-regular-tests/system-data-odbc
Current platform is: linux, linux-ppc64le, rhel, rhel-ppc64le, rhel9, rhel.9, rhel.9-ppc64le
system-data-odbc                                            [**PASS**]      (10868.3953ms)

Swapnali911:
FYI @Vishwanatha-HD

omajid:
Essentially, trying to make sure `System.Data.Odbc` and related database features work. They were found to be broken in 7.0.100 on `ppc64le` and we didn't have any other test catching it.

This test has also identified that RID parsing is still broken in `alpine:edge`: https://github.com/redhat-developer/dotnet-regular-tests/pull/222#issuecomment-1382256145

tmds:
> They were found to be broken in 7.0.100 on ppc64le and we didn't have any other test catching it.

Was it also broken due to a rid issue?

omajid:
> Was it also broken due to a rid issue?

Yes. That was because of missing `runtimes` section in `Microsoft.NETCore.App.deps.json`.

tmds:
I think we can come up with smaller, less complex tests to detect rid issues.

omajid:
Yeah, that makes sense. Do you have any particular ideas? We can definitely add simpler tests to verify RIDs - and we already have some of those, like https://github.com/redhat-developer/dotnet-regular-tests/pull/220.

That said, I think there's some value in making at least some database operations are working - since there's little other testing performed on a source-build SDK to make sure it can use databases. Many of our users going to be using some sort of database (probably through EFCore instead of System.Data.Odbc, I hope) and validating that this scenario works is beneficial even if RIDs were never an issue.

tmds:
imo this suite should focus on things that may go wrong during a .NET source build.

rids have always been a source of trouble.

I don't think there is anything specific about `System.Data.Odbc` which would cause it to break on a source-build vs Microsoft build of .NET.

To functionally validate `System.Data.Odbc` we run the functional test suite that is part of the runtime repo.

I don't expect this test to capture something that is not covered by those tests. If it would, we should look at updating the runtime tests first, so issues/regressions are detected as part of the upstream development.

Issue #238, opened 2023-03-15
Re-enable debugging-sos-lldb-via-core


tmds:
The test was disabled in https://github.com/redhat-developer/dotnet-regular-tests/pull/138

If there are some parts to the test that no pass due to upstream issues, we should leave them out.

PR #240, opened 2023-03-17
Re-enable lldb debug test.

tmds:
Fixes https://github.com/redhat-developer/dotnet-regular-tests/issues/238.

@omajid ptal.


tmds:
I don't think the test I'm changing here runs in CI due to:

https://github.com/redhat-developer/dotnet-regular-tests/blob/ef6fe84ad6b7607d7a3b1045e463a27c332ed53e/.github/workflows/ci.yml#L81-L82

So the CI failures are unrelated.

On Fedora 37, with these changes, for running this test I get: `PASS`.

omajid:
Thanks!

PR #243, opened 2023-04-05
Update debugging-via-dotnet-dump to work with upstream changes

omajid:
The implementation of dotnet-dump is being worked on, which has resulted in some changes to the information it displays. Update the test to work against it too.

### DIFF
--- a/debugging-via-dotnet-dump/test.sh
+++ b/debugging-via-dotnet-dump/test.sh
@@ -89,6 +89,8 @@ heading "Installing dotnet-dump"
 
 dotnet tool uninstall -g dotnet-dump || true
 dotnet tool install -g dotnet-dump
+# For preview releases:
+#dotnet tool install -g dotnet-dump --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
 
 framework_dir=$(../dotnet-directory --framework "${sdk_version}")
 test -f "${framework_dir}/createdump"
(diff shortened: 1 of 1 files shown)


debugging-via-dotnet-dump/test.sh
```diff
@@ -453,10 +455,10 @@ cat dump.out
 # Skipped on aarch64 due to https://github.com/dotnet/diagnostics/issues/3726
 if [[ "$(uname -m)" != "aarch64" ]]; then
     id=$(grep -F 'System.Threading.Tasks' dump.out | head -1 | cut -d' ' -f 2)
-    dump-analyze "gcroot -all ${id}" > dump.out
```
omajid, line 456: The `-all` option no longer exists in the new versions. It doesn't seem to make a practical difference in terms of things we test, though.

debugging-via-dotnet-dump/test.sh
```diff
@@ -89,6 +89,8 @@ heading "Installing dotnet-dump"
 
 dotnet tool uninstall -g dotnet-dump || true
 dotnet tool install -g dotnet-dump
+# For preview releases:
+#dotnet tool install -g dotnet-dump --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
```
aslicerh, line 93: Maybe this can be done if an ENV var is set? Something so that it can b enabled/disabled without having to change the code.

debugging-via-dotnet-dump/test.sh
```diff
@@ -554,7 +556,17 @@ cat dump.out
 id=$(tail -1 dump.out | cut -d' ' -f 2)
 dump-analyze "objsize ${id}" > dump.out
 cat dump.out
-grep -E '^sizeof\([0-9a-fA-F]+\) += +[0-9]+ \(0x[0-9a-fA-F]+\) bytes' dump.out
+if grep -F 'transitively keep alive' dump.out; then
```
aslicerh, line 559: Is it just a version change that causes this difference, or something about the running env that gets dumped? I ask because checking for aversion number or something might better express why this is here.

debugging-via-dotnet-dump/test.sh
omajid: Technically, this addition is not even necessary: the test runner injects this source (where preview versions are available) when running this test anyway.

It's only when you are debugging this script itself (without using the test runner) that this additional source plays a part. And at that point, you are probably changing the code too. Reading/enabling the env var about be about as much effort, it seems to me.

 Would you be okay with leaving it as is? Perhaps I should remove this completely?

debugging-via-dotnet-dump/test.sh
omajid: The `dotnet dump` tool is undergoing significant development (essentially, a rewrite from C++ to C#), resulting in a number of recent changes. The output change is a result of that: older versions printed the object size, the newer versions also show method tables and other information. The changes do bump the version number of the tool ([`7.0.410101`](https://www.nuget.org/packages/dotnet-dump) -> `7.0.418102`).

Checking for the this patch-version boundry seemed too brittle and so I decided to check the output instead.

debugging-via-dotnet-dump/test.sh
omajid: `-all` doesn't exist as an option; it's the default in new versions.

omajid:
Received a comment offline that the issues aren't blockers, and I really need to get this fix in for something I am working on, so I am merging this.

@aslicerh Thanks for the review! Please feel free to continue the discussion and/or file issues if you think my explanations aren't sufficient and we should further improve things.

Issue #235, opened 2023-03-07
Support running against vmr output


tmds:
Some of these tests are meant to be ran against a released version of .NET that is installed on the system.

It would be nice if we add a way of skipping those, so the remaining tests can be used in daily runs against the vmr output.

cc @omajid

omajid:
If tests can be run against the VMR output, perhaps they should be added to the VMR test suite itself? That would shift testing further left and reduce surprises as we get closer to the release.

I don't know if that's doable for all/many of our tests, though. Some of them get pretty OS/environment specific. So this would be quite nice to have.

PR #252, opened 2023-06-13
Add skipWhen=ci for tests that are known to fail on a vmr ci build.

tmds:
Fixes https://github.com/redhat-developer/dotnet-regular-tests/issues/235.

@omajid ptal.

### DIFF
--- a/apphost-framework-lookup/test.json
+++ b/apphost-framework-lookup/test.json
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "ci" // portable runtime packs not published
+  ],
   "ignoredRIDs":[
     "linux-s390x",
     "linux-ppc64le"
--- a/bash-completion/test.json
+++ b/bash-completion/test.json
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "ci" // bash completion script not installed
+  ],
   "ignoredRIDs":[
   ]
 }
--- a/bundled-libunwind/test.json
+++ b/bundled-libunwind/test.json
@@ -6,6 +6,10 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "os=fedora,version=6", // .NET 6 on Fedora uses system libunwind.
+    "os=fedora,version=7"  // .NET 7 on Fedora uses system libunwind.
+  ],
   "ignoredRIDs":[
     "alpine",
     "fedora",
--- a/debugging-via-dotnet-dump/test.sh
+++ b/debugging-via-dotnet-dump/test.sh
@@ -3,9 +3,8 @@
 # Check whether dotnet-dump, and its subcommands like `ps`, `collect`
 # and `analyze` are working
 
-if [ -f /etc/profile ]; then
-  source /etc/profile
-fi
+# Ensure global tools are on PATH.
+export PATH=~/.dotnet/tools:$PATH
 
 # Enable "unofficial strict mode" only after loading /etc/profile
 # because that usually contains lots of "errors".
--- a/omnisharp/test.json
+++ b/omnisharp/test.json
@@ -5,6 +5,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "version=8" // see https://github.com/redhat-developer/dotnet-regular-tests/issues/250
+  ],
   "ignoredRIDs":[
     "linux-musl",
     "linux-s390x",
--- a/system-libunwind/test.json
+++ b/system-libunwind/test.json
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "os=fedora,version=8" // .NET 8 on Fedora uses bundled libunwind.
+  ],
   "ignoredRIDs":[
     "centos.8",
     "centos.9",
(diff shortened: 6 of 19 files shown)


apphost-framework-lookup/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "ci" // portable runtime packs not published
```
omajid, line 10: The packages are probably getting published to an development-only feed, right? Should we somehow find and add that so we can run more of these tests?

bash-completion/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "ci" // bash completion script not installed
```
omajid, line 10: Is `ci` the right term to mean "non-packaged builds"? It is possible (and a distant goal of mine) that we can package up nightly commit of .NET (now VMR) as an RPM package and test it.

bundled-libunwind/test.json
```diff
@@ -6,6 +6,10 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "os=fedora,version=6", // .NET 6 on Fedora uses system libunwind.
+    "os=fedora,version=7"  // .NET 7 on Fedora uses system libunwind.
+  ],
```
omajid, line 12: This is now duplicating `ignoredRIDs` below. That list already contains Fedora so it would skip this test on all versions of Fedora already.

bash-completion/test.json
tmds: I'm open to any other name here. This is the one that came to mind.

omnisharp/test.json
```diff
@@ -5,6 +5,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "version=8" // see https://github.com/redhat-developer/dotnet-regular-tests/issues/250
```
omajid, line 9: Could you remove https://github.com/redhat-developer/dotnet-regular-tests/blob/43c34b5608b0957301a03b79b77b832084db8328/omnisharp/test.sh#L22-L25 and add it here too?

debugging-via-dotnet-dump/test.sh
```diff
@@ -3,9 +3,8 @@
 # Check whether dotnet-dump, and its subcommands like `ps`, `collect`
 # and `analyze` are working
 
-if [ -f /etc/profile ]; then
-  source /etc/profile
-fi
+# Ensure global tools are on PATH.
+export PATH=~/.dotnet/tools:$PATH
```
omajid, line 7: Note to self: we should a test to make sure /etc/profile adds dotnet tools to path.

apphost-framework-lookup/test.json
tmds: I don't plan to look into this as part of this PR.

bundled-libunwind/test.json
tmds: It seems that .NET 8 behavior is different? See https://github.com/redhat-developer/dotnet-regular-tests/pull/252#discussion_r1228021274.

omnisharp/test.json
tmds: #done

system-libunwind/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "os=fedora,version=8" // .NET 8 on Fedora uses bundled libunwind.
```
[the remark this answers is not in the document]
tmds: This test checks if libcoreclr.so is linked against libunwind.

With Fedora .NET 8 preview 4 packages:
```
$ ldd /usr/lib64/dotnet/shared/Microsoft.NETCore.App/8.0.0-preview.4.23259.5/libcoreclr.so
	linux-vdso.so.1 (0x00007ffd5ffb3000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f4cfabcc000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f4cfa800000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f4cfaaec000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f4cfa623000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4cfb319000)
```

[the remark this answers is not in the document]
omajid: Thanks. I will track it as a bug in our packages.

bash-completion/test.json
omajid: Some other options:

- `sdk-tarball`
- `tarball`
- `not-installed`
- `not-packaged`

tmds: Which one do you prefer?

debugging-via-dotnet-dump/test.sh
omajid: https://github.com/redhat-developer/dotnet-regular-tests/pull/255

bash-completion/test.json
tmds: I used the name `ci` to mark the tests that are skipped in our ci. And with a comment why that test is skipped.
So with one name I'm describing the traits of the ci.
I can change it to `not-packaged` if you prefer that, though it means something different than `ci` then. 🙃

tmds: I've renamed it to `vmr-ci`.

tmds:
CI failures are due to the test runner failing on the omnisharp test because the test runner version used doesn't understand the `skipWhen` condition yet.

PR #256, opened 2023-06-13
Add a test for file descriptor exhaustion

omajid:
The mono based runtimes (used on ppc64le/s390x) have a bug where the process has a low limit for file descriptors and then exhausts them quickly. More at https://github.com/dotnet/runtime/issues/82428. This was fixed for .NET 8

Add a check for that.


tmds:
Alternative suggestion for the test implementation:

From a .NET process (like an xunit test), start a process for `ulimit -Sn` and one for `ulimit -Hn` and verify their outputs are the same.

omajid:
Closing in favour of https://github.com/redhat-developer/dotnet-regular-tests/pull/270

PR #264, opened 2023-06-21
Add tests that publish with the .NET 8 bundled runtime packs.

tmds:


omajid:
Thanks for the PR!

I am a little surprised the 8.0 tests (eg https://github.com/redhat-developer/dotnet-regular-tests/actions/runs/5330802251/jobs/9657871653?pr=264) are passing. Those tests currently install 8.0 Preview 5, which is known to not include the bundled runtime packs. Does dotnet simply fall back to the packages on nuget.org if the bundled packages are missing? If so, should we set up a custom a nuget.config to clear all nuget sources?

tmds:
For an official release, this will pass due to the fallback to Microsoft `linux-{arch}` packages (except on s390x and ppc64le).
I'll add a nuget config to prevent that.

omajid:
Thanks, the new failures look much better ...

Issue #265, opened 2023-06-21
RHEL 7 doesn't support `systemd-run -E`


omajid:
https://github.com/redhat-developer/dotnet-regular-tests/blob/effd76ffe98dc51d314be036ad32a43926c46906/cgroup-limit/test.sh#L37

On RHEL 7:

```
+ mapfile -t DOTNET_LIMITS
++ sudo -n systemd-run -E PATH=/opt/rh/llvm-toolset-12.0/root/usr/bin:/opt/rh/llvm-toolset-12.0/root/usr/sbin:/opt/rh/rh-nodejs10/root/usr/bin:/opt/rh/rh-python36/root/usr/bin:/opt/rh/rh-dotnet60/root/usr/bin:/opt/rh/rh-dotnet60/root/usr/sbin:/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/tester/.local/bin:/home/tester/bin:/home/tester/.dotnet/tools -E DOTNET_ROOT=/opt/rh/rh-dotnet60/root/usr/lib64/dotnet -q --scope -p CPUQuota=100% -p MemoryLimit=100M bin/Release/net6.0/cgroup-limit
systemd-run: invalid option -- 'E'
```

tmds:
This regressed in https://github.com/redhat-developer/dotnet-regular-tests/pull/258.

PR #266, opened 2023-06-22
cgroup-limit: only pass DOTNET_ROOT to systemd-run when it is set.

tmds:
Fixes https://github.com/redhat-developer/dotnet-regular-tests/issues/265.

### DIFF
--- a/cgroup-limit/test.sh
+++ b/cgroup-limit/test.sh
@@ -4,6 +4,8 @@ set -euo pipefail
 
 set -x
 
+runtime_id="$(../runtime-id)"
+
 cat /proc/self/mountinfo
 
 cat /proc/self/cgroup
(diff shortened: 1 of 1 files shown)


cgroup-limit/test.sh
```diff
@@ -34,7 +34,10 @@ SYSTEMD_RUN="systemd-run"
 if [ "$UID" != "0" ]; then
   if ! grep -q "cpu" "/sys/fs/cgroup/user.slice/user-$UID.slice/user@$UID.service/cgroup.controllers" ; then
     # user can't set cpu limits, use sudo.
-    SYSTEMD_RUN="sudo -n $SYSTEMD_RUN -E PATH=$PATH -E DOTNET_ROOT=$DOTNET_ROOT"
+    SYSTEMD_RUN="sudo -n $SYSTEMD_RUN"
+
+    # Pass DOTNET_ROOT to support testing against a dotnet tarball.
+    [ -z "${DOTNET_ROOT}" ] || SYSTEMD_RUN="$SYSTEMD_RUN -E DOTNET_ROOT=$DOTNET_ROOT"
```
omajid, line 40: I am not sure I follow this change.

`DOTNET_ROOT` is defined on RHEL 7 when we are testing our packages. It's set here: https://git.centos.org/rpms/rh-dotnet60/blob/c7-dotnet/f/SPECS/rh-dotnet60.spec#_66. So that still results in us trying to run `systemd-run -E` which fails as expected.

tmds: > DOTNET_ROOT is defined on RHEL 7 when we are testing our packages.

Can you remind me why we're setting `DOTNET_ROOT`? Don't we have the root in `/etc/dotnet/install_location`?

tmds: Probably because of the scl.

I'll change the test to not add the arg on rhel7.

PR #270, opened 2023-07-07
Add a test to make sure file descriptor limits are correct

omajid:
Some versions of the mono based runtimes (used on ppc64le/s390x) are know to have a bug where the process has a low limit for file descriptors and then exhausts them quickly. It's easy to run into this if you use many nuget sources, for example. More at dotnet/runtime#82428. This was fixed for .NET 8

### DIFF
--- a/limits/LimitsTest.cs
+++ b/limits/LimitsTest.cs
@@ -1,0 +1,41 @@
+using System.Diagnostics;
+using System.Globalization;
+using Xunit;
+
+namespace Limit;
+
+public class LimitsTest
+{
+    [Fact]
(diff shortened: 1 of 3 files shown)


limits/LimitsTest.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            FileName = name,
+            RedirectStandardOutput = true,
+        };
+        foreach (string arg in args)
+        {
+            psi.ArgumentList.Add(arg);
+        }
+
+        Process? p = Process.Start(psi);
```
tmds, line 30: Fyi, on Linux this never returned null.

PR #271, opened 2023-07-10
Fix cgroup-limit to work on RHEL 7

omajid:
runtime_id is an RID and looks like `rhel.7-x64`, so it needs to be tested correctly.


omajid:
This is a followup to https://github.com/redhat-developer/dotnet-regular-tests/pull/266

FYI @tmds

tmds:
oops ... thanks for fixing!

Maybe we can extend CI here so it would have caught this, and https://github.com/redhat-developer/dotnet-bunny/pull/79?

PR #279, opened 2023-08-25
Add a test for testing dynamic PGO

omajid:
This was initially added to .NET 7, make sure it works correctly for .NET 7 and later.


omajid:
TODO: Check that PGO is enabled by default for .NET 8 and later: https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#performance-improvements

Issue #259, opened 2023-06-16
debugging-sos-lldb-via-core fails on RHEL9 with .NET 8


tmds:
Failure on RHEL9 x64 .NET8 vmr preview build:
```
+ lldb --batch -c coredump.193103 --one-line 'gcroot 7fb2f8839f28'
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: lldb --batch -c coredump.193103 --one-line "gcroot 7fb2f8839f28"
1.	HandleCommand(command = "gcroot 7fb2f8839f28")�
 #0 0x00007f01da3a8e45 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/lib64/libLLVM-15.so+0xba8e45)
 #1 0x00007f01da3a6ee4 llvm::sys::RunSignalHandlers() (/lib64/libLLVM-15.so+0xba6ee4)
 #2 0x00007f01da3a933d (/lib64/libLLVM-15.so+0xba933d)
 #3 0x00007f01c76989ae IsRunningOnAlternateStack(void*) /home/tester/dotnet/src/runtime/artifacts/source-build/self/src/src/coreclr/pal/src/exception/signal.cpp:318:9
 #4 0x00007f01c76989ae invoke_previous_action(sigaction*, int, siginfo_t*, void*, bool) /home/tester/dotnet/src/runtime/artifacts/source-build/self/src/src/coreclr/pal/src/exception/signal.cpp:414:31
 #5 0x00007f01c7697e97 sigsegv_handler(int, siginfo_t*, void*) /home/tester/dotnet/src/runtime/artifacts/source-build/self/src/src/coreclr/pal/src/exception/signal.cpp:0:5
 #6 0x00007f01d9054df0 __restore_rt (/lib64/libc.so.6+0x54df0)
 #7 0x00007ec067f37c05 ReadyToRun_EnclosingTypeMap::GetEnclosingTypeNoThrow(unsigned int, unsigned int*, IMDInternalImport*) const /home/tester/dotnet/src/runtime/artifacts/source-build/self/src/src/coreclr/vm/readytoruninfo.cpp:1823:16
[...]

PR #280, opened 2023-08-29
debugging-sos-lldb-via-core: re-enable test for .NET 8.

tmds:
The issues that caused the test to fail have been fixed.

Also remove the debugging-sos-lldb-via-core-2x variant which implemented the test for .NET Core 2.x.

Closes https://github.com/redhat-developer/dotnet-regular-tests/issues/259.

cc @omajid


tmds:
> The issues that caused the test to fail have been fixed.

I validated this locally using an 8.0.100-rc.2.23428.1 build on Fedora x64.

PR #302, opened 2023-09-28
adding test coverage for dotnet-trace.

### LINKED ISSUE #223: Add a test to verify `dotnet-trace` works
We should make sure that https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-trace works against every  build of .NET

nicrowe00:
Adding test coverage for the diagnostic tool 'dotnet-trace'. This test covers the basic functions of dotnet-trace to ensure it ran correctly. All of the main commands used in dotnet-trace are covered, the commands being collect, ps, list-profiles, convert and report.

Closes #223

### DIFF
--- a/dotnet-trace/test.sh
+++ b/dotnet-trace/test.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+set -x
+
+PROJNAME=dotnettrace
+FILENAME=dotnettrace.nettrace
+SPEEDSCOPENAME=dotnettrace.speedscope.json
+REPORTNAME=dotnettrace.nettrace.etlx
+
+dotnet tool update -g dotnet-trace
+export PATH="$HOME/.dotnet/tools:$PATH"
+
+dotnet-trace collect -o $FILENAME -- dotnet new console --output $PROJNAME
+if [ -f $FILENAME ]; then
+   echo "collect - OK"
+else
+   echo "collect - FAIL"
+   rm -r $PROJNAME
+   exit 1
+fi
+
+if dotnet-trace ps; then
+   echo "ps - OK"
+else
+   echo "ps - FAIL"
+   rm -r $PROJNAME
+   rm $FILENAME
+   exit 1
+fi
+
+if dotnet-trace list-profiles; then
+   echo "list-profiles - OK"
+else
+   echo "list-profiles - FAIL"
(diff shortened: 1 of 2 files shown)


nicrowe00:
@omajid @tmds PTAL when you have a chance.

dotnet-trace/test.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+PROJNAME=dotnettrace
+FILENAME=dotnettrace.nettrace
+SPEEDSCOPENAME=dotnettrace.speedscope.json
+REPORTNAME=dotnettrace.nettrace.etlx
+
+
```
omajid, line 10: Lets make the test install the tool dependency before using it.

```suggestion
dotnet tool uninstall -g dotnet-trace || true
dotnet tool install -g dotnet-trace

```

tmds: You can use `dotnet tool update` instead.

If the tool isn't installed yet, it will install it.
If the tool is installed, it will update it to the latest version.

dotnet-trace/test.sh
```diff
@@ -0,0 +1,65 @@
...
+set -euo pipefail
+set -x
+
+PROJNAME=dotnettrace
+FILENAME=dotnettrace.nettrace
+SPEEDSCOPENAME=dotnettrace.speedscope.json
+REPORTNAME=dotnettrace.nettrace.etlx
+
+dotnet tool update -g dotnet-trace > /dev/null
```
omajid, line 11: I think redirecting output to /dev/null means we lose command and debugging output. What do you think about leaving that in? That way, when something fails, we can at least see what the output of the command was? And even when things pass, we can see, for example, what version of the tool was installed.

nicrowe00: Sounds good to me, I'll change it now.

dotnet-trace/test.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ -f $SPEEDSCOPENAME ]; then
+   echo "convert - OK"
+else
+   echo "convert - FAIL"
+   rm -r $PROJNAME
+   rm $FILENAME
+   exit 1
+fi 
+
+dotnet-trace report $FILENAME topN
```
tmds, line 51: It would be nice to check if something expected is in the trace file.

PR #304, opened 2023-10-02
follow-up for dotnet-trace

nicrowe00:
Rewriting a portion of the test for dotnet-trace, specifically dealing with the 'report' command. Instead of verifying that the report log file exists, the test now checks the output of the command to verify that a specific method is on the callstack.

### DIFF
--- a/dotnet-trace/test.json
+++ b/dotnet-trace/test.json
@@ -2,7 +2,7 @@
 	"name": "dotnet-trace",
 	"enabled": true,
 	"requireSdk": true,
-	"version": "6.0",
+	"version": "8.0",
 	"versionSpecific": false,
 	"type": "bash",
 	"cleanup": false,
--- a/dotnet-trace/test.sh
+++ b/dotnet-trace/test.sh
@@ -48,14 +48,22 @@ else
    exit 1
 fi 
 
-dotnet-trace report $FILENAME topN
-if [ -f $REPORTNAME ]; then
-   echo "report - OK"
-else 
+output=$(dotnet-trace report $FILENAME topN)
+if [[ $output == *"Missing Symbol"* ]]; then 
    echo "report - FAIL"
    rm -r $PROJNAME
    rm $FILENAME
    rm $SPEEDSCOPENAME
+   rm $REPORTNAME
+   exit 1
+elif [[ $output == *"Top 5 Functions (Exclusive)"* ]]; then
+   echo "report - OK"
+else
+   echo "report - FAIL"
+   rm -r $PROJNAME
+   rm $FILENAME
+   rm $SPEEDSCOPENAME
+   rm $REPORTNAME
    exit 1
 fi
 


nicrowe00:
@omajid @tmds PTAL

dotnet-trace/test.sh
```diff
@@ -48,14 +48,14 @@ else
    exit 1
 fi 
 
-dotnet-trace report $FILENAME topN
-if [ -f $REPORTNAME ]; then
+if dotnet-trace report $FILENAME topN | grep -q 'Missing Symbol'; then
```
omajid, line 51: `grep -q` exits immediately if a match is found, which can cause the previous commands in the pipeline to report a SIGPIPE. That leads to non-reproducible and hard to debug issues.

dotnet-trace/test.sh
```diff
@@ -48,14 +48,14 @@ else
    exit 1
 fi 
 
-dotnet-trace report $FILENAME topN
```
omajid, line 51: I think it would be nice for debugging if the entire output was written to stdout.

dotnet-trace/test.sh
omajid, line 51: Is this saying that `Missing Symbols` is expected in the output? To check that `dotnet trace report topN` is working correctly, we should check that there are no errors about missing symbols.

nicrowe00: Yes, `Missing Symbol` is expected in the output in order to verify that the command is returning the methods that have been on the callstack.

omajid: This doesn't look like the case for .NET 8.

The CI job for 8 is failing with this in the logs:

```
+ grep -q 'Missing Symbol'
+ dotnet-trace report dotnettrace.nettrace topN
+ echo 'report - FAIL'
+ rm -r dotnettrace
report - FAIL
+ rm dotnettrace.nettrace
+ rm dotnettrace.speedscope.json
+ rm dotnettrace.nettrace.etlx
+ exit 1
```

Running the topN command on my machine using .NET 8, I see output like:

```
$ dotnet trace report /home/omajid/temp/dotnet_20231002_101421.nettrace topN
Top 5 Functions (Exclusive)                                                  Inclusive           Exclusive           
1. LowLevelLifoSemaphore.WaitNative(class Microsoft.Win32.SafeHandles.Saf    57.28%              57.28%              
2. Monitor.Wait(class System.Object,int32)                                   12.94%              12.94%              
3. WaitHandle.WaitOneNoCheck(int32)                                          10.71%              10.71%              
4. WaitHandle.WaitMultiple(value class System.ReadOnlySpan`1<class System    9.04%               9.04%               
5. WaitHandle.WaitAnyMultiple(value class System.ReadOnlySpan`1<class Mic    2.43%               2.43%               
```

While .NET 7 shows:

```
$ dotnet trace report /home/omajid/temp/dotnet_20231002_112832.nettrace topN
Top 5 Functions (Exclusive)                                                  Inclusive           Exclusive           
1. Missing Symbol                                                            43.04%              42.81%              
2. WaitHandle.WaitOneNoCheck(int32)                                          10.62%              10.62%              
3. ManualResetEventSlim.Wait(int32,value class System.Threading.Cancellat    9.69%               9.69%               
4. __Canon].MoveNext()                                                       7.37%               7.33%               
5. MSBuildEvaluator.EvaluateProject(class Microsoft.TemplateEngine.Abstra    4.35%               2.21%
```

I actually think the .NET 7 output showing "Missing symbol" is a bug - it show show the symbol (method) name, not that it can't find the symbol

nicrowe00: Ah OK, thank you for clarifying this.

omajid: Digging some more (eg, https://github.com/dotnet/diagnostics/issues/3900, https://github.com/dotnet/diagnostics/blob/f4e915c7e9b5869d9b3b646aadf87755061c6638/src/Tools/dotnet-trace/CommandLine/PrintReportHelper.cs#L124), it looks like this message is printed when the debug symbols are missing. The symbols can be missing due to runtime or dotnet-trace bugs or because application-specific debug symbols are simply not installed.

With .NET 7, we didn't ship managed debug symbols as part of the SDK. So that could be an issue. Or it could be a bug in runtime or dotnet-trace. With .NET 8, we are shipping debug symbols.

How about this: let's limit this test to .NET 8 and make it fail if we see any "Missing Symbol" in the output? Then we can see if/when this test fails and figure out if it was a breakage in the runtime or dotnet-trace or something else?

dotnet-trace/test.json
```diff
@@ -2,8 +2,8 @@
 	"name": "dotnet-trace",
 	"enabled": true,
 	"requireSdk": true,
-	"version": "6.0",
-	"versionSpecific": false,
+	"version": "8.0",
+	"versionSpecific": true,
```
tmds, line 6: You should keep `versionSpecific` set to `false` so the test also runs on versions higher than .NET 8.

dotnet-trace/test.sh
```diff
@@ -48,15 +48,16 @@ else
...
-else 
+output=$(dotnet-trace report $FILENAME topN)
+if [[ $output == *"Missing Symbol"* ]]; then 
    echo "report - FAIL"
    rm -r $PROJNAME
    rm $FILENAME
    rm $SPEEDSCOPENAME
+   rm $REPORTNAME
    exit 1
+else
```
tmds, line 59: Can you validate the expected output in the OK case?

dotnet-trace/test.json
nicrowe00: Noted.

dotnet-trace/test.sh
nicrowe00: Validation has been added.

dotnet-trace/test.json
tmds: You can read the documentation for these properties on the README at https://github.com/redhat-developer/dotnet-bunny.

tmds:
@nicrowe00 fyi, the trace files generated by the tool can be read using https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/. There's an example in this blog post that shows how to use the library for reading a trace file: https://developers.redhat.com/blog/2019/12/23/tracing-net-core-applications.

I'm ok with validating the output of the topN because that will read the trace to create a summary.

omajid (approved):
LGTM!

PR #309, opened 2023-10-12
Use TestTargetFramework to match the framework under test.

tmds:
So far the test runner was patching csproj files that were directly below the test directory and named like the test directory.

With the latest test runner release, the target framework is provided as an environment variable. The environment variable makes it possible to use the target framework anywhere in the tests.

It also reduces the noise on 'git diff' as the csproj files will no longer be changed by the test runner.

This change adopts the new environment variable in the existing tests.

### DIFF
--- a/aspnet-same-runtime-version-2x/test.sh
+++ b/aspnet-same-runtime-version-2x/test.sh
@@ -30,7 +30,7 @@ function create_project_for_package()
   cat >$test_folder/testapp.csproj <<EOF
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFramework>netcoreapp${runtime_version_major_minor}</TargetFramework>
+    <TargetFramework>\$(TestTargetFramework)</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="$package" />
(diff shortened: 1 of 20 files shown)


nativeaot/console.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>$(TestTargetFramework)</TargetFramework>
```
tmds, line 5: This project didn't actually meet the naming conventions for the test runner to patch it.

aspnet-same-runtime-version-2x/test.sh
```diff
@@ -30,7 +30,7 @@ function create_project_for_package()
   cat >$test_folder/testapp.csproj <<EOF
 <Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFramework>netcoreapp${runtime_version_major_minor}</TargetFramework>
+    <TargetFramework>\$(TestTargetFramework)</TargetFramework>
```
tmds, line 33: It would be good to remove old tests like this one, which only run on versions that are no longer supported.

tmds:
The `template-test` failed on the CI runs against 6.0. The test wasn't changed by this PR. The log indicates nuget.org connection issues.

nicrowe00:
LGTM

PR #318, opened 2023-11-17
test coverage for dotnet-monitor

nicrowe00:
Adding test coverage for the diagnostic tool 'dotnet-monitor'.


dotnet-monitor/test.json
```diff
@@ -0,0 +1,11 @@
+{
```
omajid, line 1: Minor nit: I would avoid naming directory/project with the same name as something that exists on nuget.org. It's probably fine for bash scripts, but can fail surprisingly for C# code. And maybe make the directory name a bit more descriptive? Perhaps something along the lines of `dotnet-monitor-works`?

PR #319, opened 2023-11-17
specify 'dotnet-ef' version for tools-in-path

nicrowe00:
With the release of .NET 8, the tools for .NET have also been updated for .NET 8. The new updated tools are incompatiable with .NET 6 and 7 meaning the correct tool version must be specified for the version of .NET installed on the machine.

### DIFF
--- a/tools-in-path/test.sh
+++ b/tools-in-path/test.sh
@@ -7,7 +7,7 @@ if [ -f /etc/profile ]; then
 fi
 
 set -euo pipefail
-
+IFS=$'\n\t'
 set -x
 
 echo "$PATH"
@@ -19,10 +19,7 @@ else
     exit 1
 fi
 
-if dotnet tool list -g | grep -F dotnet-ef; then
-    true
-else
-    dotnet tool install -g dotnet-ef || true
-fi
+IFS='.-' read -ra VERSION_SPLIT <<< "$1" 
+dotnet tool update -g dotnet-ef --version "${VERSION_SPLIT[0]}.*-*"
 
 dotnet ef --version


nicrowe00:
The test for .NET 8 on Fedora 38 is failing as the new release of .NET 8 still isn't available in the package manager for Fedora which causes it to use .NET 8-rc2 which is incompatible with the new update for 'dotnet-ef'. This test will continue to fail until new release of .NET 8 is added to the Fedora package manager.

nicrowe00:
@tmds @omajid PTAL

tmds:
Can you try what happens if we replace the whole `if .. fi` block with:
```sh
IFS='.-' read -ra VERSION_SPLIT <<< "$1" 

dotnet tool update -g dotnet-ef --version "${VERSION_SPLIT[0]}.*-*"
```

tmds:
> The new updated tools are incompatiable with .NET 6 and 7

For reference: https://github.com/dotnet/sdk/issues/36904

tools-in-path/test.sh
```diff
@@ -19,10 +19,13 @@ else
     exit 1
 fi
 
+version=$(dotnet --version)
```
tmds, line 22: Can you replace everything including this version variable up to the `fi` with these two lines:
```
IFS='.-' read -ra VERSION_SPLIT <<< "$version"
dotnet tool update -g dotnet-ef --version "${VERSION_SPLIT[0]}.*-*"
```

If the `|| true` is needed, it would be nice if we add a comment that explains why.

tmds: Note: `$1` is set by the test runner to the runtime version.

nicrowe00: > Note: `$1` is set by the test runner to the runtime version.

Thank you, I wasn't aware of this.

omajid (approved):
Good catch!

tools-in-path/test.sh
```diff
@@ -7,7 +7,7 @@ if [ -f /etc/profile ]; then
 fi
 
 set -euo pipefail
-
+IFS=$'\n\t'
```
tmds, line 10: Why did you add this?

omajid: I assume because it's generally added next to `set -euo pipefail` as part of ["bash strict mode"](http://redsymbol.net/articles/unofficial-bash-strict-mode/)?

tmds: Should we be adding it to all tests then?

omajid: I would be okay with that. Doesn't seem very high priority though.

PR #320, opened 2023-11-28
set tests to run in bash strict mode

nicrowe00:
As discussed here: https://github.com/redhat-developer/dotnet-regular-tests/pull/319#discussion_r1397638584 , applicable tests are now being set to run bash strict mode in order to increase maintainability and reliability. More information on bash strict mode can be found here: http://redsymbol.net/articles/unofficial-bash-strict-mode/

### DIFF
--- a/aspnet-same-runtime-version/test.sh
+++ b/aspnet-same-runtime-version/test.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -euo pipefail
+IFS=$'\n\t'
 
 # Check Microsoft.AspNetCore.App's version matches that of Microsoft.NETCore.App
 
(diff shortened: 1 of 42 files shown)


aspnet-same-runtime-version/test.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -euo pipefail
+IFS=$'\n\t'
```
tmds, line 4: Let's consistently use the exact same lines as we have here in all scripts:
```
#!/usr/bin/env bash

set -euo pipefail
IFS=$'\n\t'

```
For scripts that `set -x`, let's add it as a line directly after the `IFS` assignment.

PR #322, opened 2023-12-13
Add rsa-pkcs-openssl test.

tmds:
cc @tomato42


nicrowe00 (approved):

tomato42 (changes requested):
Since implicit rejection algorithm is deterministic, the test should expect specific message returned for a given static ciphertext and static key.

tmds:
> Since implicit rejection algorithm is deterministic, the test should expect specific message returned for a given static ciphertext and static key.

I've updated the test.

tomato42 (approved):
Verified to be an invalid ciphertext and the synthethic ciphertext to match the algorithm specified in https://datatracker.ietf.org/doc/html/draft-kario-rsa-guidance-02#name-depadding

Looks good!

tmds:
Thanks for reviewing, @nicrowe00 and @tomato42!

PR #324, opened 2023-12-15
Refractoring cgroup-limit

nicrowe00:
Removing the line from cgroup-limit that was enabling bash strict mode as it was causing CI failures.


omajid:
FYI, since this is intentionally changing what the code does, we shouldn't call this refactoring. Refactoring is changing the organization/layout/structure/execution of the code **without** change the observable behaviour.

PR #329, opened 2024-01-23
Accept a version string for .NET 9

omajid:
Tthis test fails against the in-development versions of .NET 9 because it only accepts versions up to (and including) 8.


omajid:
CI errors look unrelated.

nicrowe00 (approved):

PR #332, opened 2024-01-26
template-test: consider armv8l as arm

ayakael:


ayakael:
Current failure has to do with removing `libcoreclrtraceptprovider.so` due https://github.com/dotnet/runtime/issues/57784. @omajid Is best practice still to remove that file since lttnng-ust tracing is still broken?

omajid:
On Fedora/RHEL, we leave that file around. It doesn't seem to hurt anything as far as I can tell.

Also on Fedora/RHEL, with newer versions of .NET, we have made the dependency on lttng-ust optional at the distro-package-level. `libcoreclrtraceptprovider.so` correctly handles the case when lttng-ust is missing from the distro.

lttng/test.json
```diff
@@ -12,6 +12,7 @@
   "ignoredRIDs":[
     "centos",
     "fedora", // see https://github.com/redhat-developer/dotnet-regular-tests/issues/202
+    "alpine", // lttng tracing disabled due to no support for lttng-ust 2.13
```
omajid, line 15: If there's an issue on github/gitlab/jira/whatever with details, linking to that would be nice too.

ayakael:
I dropped the lttng changes as I adopted your approach to dealing with this bug. The test units passes now.

omajid:
Thanks!

PR #334, opened 2024-01-29
Extend distribution-packages to cover -dbg packages too

omajid:


omajid:
There's a chicken and egg problem with getting tests to pass. I would like the tests to get enabled so I can use them when doing package builds and make sure my packaging changes are completely. On the other hand, until the packaging changes are complete, the newly-introduced tests will keep on failing.

I would like to merge this, and then fix the packages over the next few weeks. Unfortunately, that means the tests will stay failing in CI for the next few weeks.

Anyone have objections to this?

nicrowe00:
I would prefer that the test was properly tested and passing before introducing it to CI but if it's necessary for completing the package builds then I have no issue with merging it.

cc @tmds

tmds:
We should aim to minimize the time this test is causing failures.

Note: they won't affect the hydra jobs, where this test gets skipped. They will affect jobs like the ones for which we've just merged https://github.com/redhat-developer/dotnet-regular-tests/pull/333.

omajid:
> I would prefer that the test was properly tested and passing before introducing it to CI but if it's necessary for completing the package builds then I have no issue with merging it.

I tested this locally and it passes with fixed-up packages. Test pipeline in this PR shows that it correctly flags issues in packages that are missing -dbg subpackage. Is that good enough testing to make sure this test works correctly? If we find any issues, I am going to fix it up.

This PR is not a strict requirement to make the packaging changes, but it makes sure the changes are complete and working. Without this test running, if I screw up a package change, then merge this test, fixing it will take even longer, and might miss the February update window.

> We should aim to minimize the time this test is causing failures.

My goal is to have this passing by mid-Feb everywhere.

PR #340, opened 2024-02-07
Use incremental backoff instead of hardcoding fixed sleep times

omajid:
With fixed/long sleep times, tests run slower when there is no need to wait for that long. When the machine is slow (or under heavy load), we often to need to wait for longer before things are running and the fixed sleep times are not sufficient.

### DIFF
--- a/run-until-success-with-backoff
+++ b/run-until-success-with-backoff
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -euo pipefail
+
+echo "Running with backoff:" "$@"
+
+max_retries=10
+
+iterations=0
+total_slept=0 # seconds
+
+sleep $((iterations + 1))
+((total_slept += iterations + 1))
+
+until "$@"; do
+    ((iterations += 1))
+    if (( total_slept > 10 )); then
+        echo "$@" "still failing after more than ${total_slept} seconds"
+    fi
+    if (( iterations == max_retries )); then
+        echo "$@" "still failing after $max_retries retries"
+        exit 1
+    fi
+    sleep $iterations
+    ((total_slept += iterations))
+done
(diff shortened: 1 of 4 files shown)


run-until-success-with-backoff
```diff
@@ -0,0 +1,16 @@
...
+
+set -euo pipefail
+
+set -x
+
+i=0
+max_retries=20
+until "$@"; do
+    ((i+=1))
+    if [[ $i == "$max_retries" ]]; then
```
tmds, line 11: Can we start printing messages as soon as something takes longer than 10 seconds, to give some additional visibility to things that are slow.

run-until-success-with-backoff
```diff
@@ -0,0 +1,16 @@
...
+        echo "$@" "still failing after $max_retries retries"
+        exit 1
+    fi
+    sleep $i
```
tmds, line 15: Do we want to move the sleep before the first attempt so we're giving some cpu time to what we're willing to wait for here?

run-until-success-with-backoff
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+set -x
+
+i=0
+max_retries=20
```
tmds, line 8: With max_retries 20, we're going from 5 seconds to about 200s. That's x 40.
max_retries 10 brings us to about 50s. x 10 seems plenty.

run-until-success-with-backoff
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+set -x
```
tmds, line 5: Maybe we should leave this out to avoid more output in the test, and let this helper print useful messages.

run-until-success-with-backoff
omajid: Done.

run-until-success-with-backoff
omajid: Done.

run-until-success-with-backoff
omajid: Done.

run-until-success-with-backoff
omajid: Good idea. Done.

omajid:
I am merging this because the insufficient-sleep issue is making it much more difficult to get passing tests for me. If there are any outstanding concerns with this PR, please let me know and we can work through them in a separate PR. TIA.

PR #341, opened 2024-02-09
Adding aarch64, s390x and ppc64le to CI

### LINKED ISSUE #273: Add arm64, ppc64le and s390x to CI
Our tests only verify that things work on x86_64 (aka x64). We don't test on other architecture. Because of that, we run into issues and have ot fix them later. We should add some test legs for those other architectures too.

nicrowe00:
Adding aarch64,s390x and ppc64le for fedora

Closes #273

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,13 @@ jobs:
           - quay.io/centos/centos:stream8
           - quay.io/centos/centos:stream9
           - registry.fedoraproject.org/fedora:38
+          - registry.fedoraproject.org/fedora:38-aarch64
+          - registry.fedoraproject.org/fedora:38-s390x
+          - registry.fedoraproject.org/fedora:38-ppc64le
           - registry.fedoraproject.org/fedora:39
+          - registry.fedoraproject.org/fedora:39-aarch64
+          - registry.fedoraproject.org/fedora:39-s390x
+          - registry.fedoraproject.org/fedora:39-ppc64le
           - registry.fedoraproject.org/fedora:rawhide
           - registry.access.redhat.com/ubi8
           - registry.access.redhat.com/ubi9


.github/workflows/ci.yml
```diff
@@ -32,7 +44,11 @@ jobs:
           - "7.0"
           - "8.0"
         exclude:
-          - container_image: docker.io/library/alpine:latest
+          container_image: 
+            - docker.io/library/alpine:latest 
+            - docker.io/arm64v8/alpine:latest 
+            - docker.io/s390x/alpine:latest 
+            - docker.io/ppc64le/alpine:latest
```
tmds, line 51: I find this yaml confusing, but I assume it means we're skipping the 8.0 on these Alpine images? Why is that?
If we're skipping the 8.0, maybe we should just leave these images out completely?

nicrowe00: I set arm64, s390x and ppc64le to skip 8.0 because docker.io/library/alpine:latest had been already set to skip it for reasons presently unknown to me. I assumed there would be a similar issue with the other three architectures. I don't mind leaving out arm64, s390x and ppc64le for alpine for now.

omajid:
Any idea why GitHub Actions didn't run any testing legs for this PR?

omajid:
Thank you for doing this! This would be amazing to have!

nicrowe00:
> Any idea why GitHub Actions didn't run any testing legs for this PR?
> 
> Edit: It's probably invalid yaml syntax?

Yeah, it was invalid syntax that caused them not to run. Before I pushed the second time with the invalid syntax it was running but the new jobs were failing because the containers wouldn't start. I'll fix the invalid syntax so that the testing legs will run and debug why the containers won't start.

omajid:
> debug why the containers won't start.

I suspect that GitHub Actions aren't configured to run cross-architectures containers. With podman/docker, you need to use recent versions of the tool and pass special arguments to run cross-architecture contianers using emulators. We probably need to find equivalents (if possible at all) for GitHub Actions.

tmds:
> using emulator

Probably this is going to be sloooooooooow ... but we can give it a try and see what happens.

PR #345, opened 2024-03-01
Initial changes for .NET 9 RPMs

omajid:
I built .NET 9 Preview 1 locally and ran tests against it. Here are some obvious fixes.

### DIFF
--- a/distribution-packages/test-standard-packages
+++ b/distribution-packages/test-standard-packages
@@ -325,11 +325,8 @@ def main() -> int:
 
 def to_package_version(version: str) -> str:
     if '-preview.' or '-rc.' in version:
-        # Convert upstream version 11.2.3-preview.999.1234 to RPM-style 11.2.3~preview.999
-        parts = version.split('-')
-        if len(parts) == 2:
-            release_prefix = '.'.join(parts[1].split('.')[0:2])
-            return parts[0] + '~' + release_prefix
+        # Convert upstream version 11.2.3-preview.999.1234 to RPM-style 11.2.3~preview.999.1234
+        return version.replace('-', '~')
     return version
 
 
--- a/distribution-packages/test.sh
+++ b/distribution-packages/test.sh
@@ -5,7 +5,7 @@ IFS=$'\n\t'
 set -x
 
 sdk_version="$(dotnet --version)"
-runtime_version="$(dotnet --list-runtimes | head -1 | awk '{ print $2 }')"
+runtime_version="$(dotnet --list-runtimes | grep Microsoft.NETCore.App | awk '{ print $2 }')"
 runtime_id=$(../runtime-id)
 # This might be the final/only netstandard version from now on
 netstandard_version=2.1
--- a/tools-in-path/test.sh
+++ b/tools-in-path/test.sh
@@ -20,6 +20,6 @@ else
 fi
 
 framework_dir=$(../dotnet-directory --framework "$1")
-dotnet tool update -g dotnet-ef --version "${framework_dir#*App/}.*-*"
+dotnet tool update -g dotnet-ef --version "${framework_dir#*App/}.*"
 
 dotnet ef --version
--- a/template-test/test.sh
+++ b/template-test/test.sh
@@ -11,6 +11,49 @@ set -x
 # If additional templates are found via `dotnet new --list`, this test
 # will fail unless they are added here.
 
+dotnet9Templates=(
+    apicontroller
+    blazor
+    blazorwasm
+    buildprops
+    buildtargets
+    classlib
+    console
+    editorconfig
+    .editorconfig
+    gitignore
+    .gitignore
+    global.json
+    globaljson
+    grpc
+    mstest
+    mvc
+    mvccontroller
+    nuget.config
+    nugetconfig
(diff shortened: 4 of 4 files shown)


distribution-packages/test-standard-packages
```diff
@@ -325,11 +325,8 @@ def main() -> int:
 
 def to_package_version(version: str) -> str:
     if '-preview.' or '-rc.' in version:
-        # Convert upstream version 11.2.3-preview.999.1234 to RPM-style 11.2.3~preview.999
-        parts = version.split('-')
-        if len(parts) == 2:
-            release_prefix = '.'.join(parts[1].split('.')[0:2])
-            return parts[0] + '~' + release_prefix
+        # Convert upstream version 11.2.3-preview.999.1234 to RPM-style 11.2.3~preview.999.1234
+        return version.replace('-', '~')
```
omajid, line 329: RPMs previously used `preview.1` instead of `preview.1.24080.9` mostly because it was difficult to get accurate information before building things. With [release.json](https://github.com/dotnet/dotnet/releases/download/v9.0.0-preview.1.24080.9/release.json), this is easy to fix, so I fixed it.

distribution-packages/test.sh
```diff
@@ -5,7 +5,7 @@ IFS=$'\n\t'
 set -x
 
 sdk_version="$(dotnet --version)"
-runtime_version="$(dotnet --list-runtimes | head -1 | awk '{ print $2 }')"
+runtime_version="$(dotnet --list-runtimes | grep Microsoft.NETCore.App | awk '{ print $2 }')"
```
omajid, line 8: `dotnet --list-runtimes` lists all the runtimes, including ASP.NET Core. This filter was relying on a certain ordering that doesn't work with .NET 9 Preview 1.

tools-in-path/test.sh
```diff
@@ -20,6 +20,6 @@ else
 fi
 
 framework_dir=$(../dotnet-directory --framework "$1")
-dotnet tool update -g dotnet-ef --version "${framework_dir#*App/}.*-*"
+dotnet tool update -g dotnet-ef --version "${framework_dir#*App/}.*"
```
omajid, line 23: I am not sure why we previously used `*-*`. Will this change have a negative impact? Thoughts, @tmds ?

nicrowe00 (approved):

tools-in-path/test.sh
omajid: With a preview version, this was evaluated as `9.0.0-preview.1.24080.9*-*`, which doesn't match anything. But `9.0.0-preview.1.24080.9*` does.

tmds: @nicrowe00 updated this test since it was failing for .NET 8.0.1 here: https://github.com/redhat-developer/dotnet-regular-tests/pull/333.

This test is skipped in our CI, so we haven't noticed it doesn't work against preview versions.

With the March release (`8.0.3`), we should be able to go back to how this was before https://github.com/redhat-developer/dotnet-regular-tests/pull/333 because `dotnet-ef` should then run on any 8.0 runtime version.

omajid: Should I merge this change, then, or defer it to a separate PR (or never)?

omajid: @nicrowe00 can you remind us why we had to use `*-*` here?

tmds: > can you remind us why we had to use *-* here?

He didn't add it. It was already there.

Before https://github.com/redhat-developer/dotnet-regular-tests/pull/333 we prefixed this only with the major version number. So the `--version` argument was for example `7.*-*`. This means the SDK it is allowed to install a version higher than 7. It will prefer to install a stable version (e.g. `7.0.0`) but if there is no stable version, it can install a preview (e.g. `7.0.0-preview1`).

With https://github.com/redhat-developer/dotnet-regular-tests/pull/333, if you run against a released version of .NET, you get something like `8.0.1.*-*`. For the SDK this is an acceptable version pattern.

When https://github.com/redhat-developer/dotnet-regular-tests/pull/333 runs against a .NET 9 preview, the `--version` becomes something like  `9.0.0-preview.1.24080.9*-*` and this pattern doesn't make sense to the SDK.

This PR is now changing that version into `9.0.0-preview.1.24080.9.*`. And if this is working, it seems the SDK treats this the same as `9.0.0-*` (since the published `dotnet-ef` version is actually `9.0.0-preview.1.24081.2`).

> Should I merge this change, then, or defer it to a separate PR (or never)?

I'm fine with either what you have here, or the pattern we had before https://github.com/redhat-developer/dotnet-regular-tests/pull/333 (which will work for .NET 9 previews, and should start working for .NET 8 on the next patch release).

omajid: > He didn't add it. It was already there.

I was thinking of #319.

tmds: Because I suggested it in that PR for the reasons mentioned above.

tmds: This is the documentation for the wildcard pattern: https://github.com/NuGet/Home/wiki/Support-pre-release-packages-with-floating-versions.

PR #347, opened 2024-03-06
test for SHA1 CAs

nicrowe00:


sha1-validation/test.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "sha1-validation",
+    "enabled": true,
+    "requiresSdk": true,
+    "version": "8.0",
+    "versionSpecific": true,
+    "type": "xunit",
```
omajid, line 7: Should be type `bash`?

nicrowe00: Thank you.

nicrowe00:
Test failures are unrelated except for .NET 8 on UBI9 which we wanted to fail.

omajid:
Do we want to check .NET 6 on UBI9?

nicrowe00:
> Do we want to check .NET 6 on UBI9?

I tested .NET 6 on UBI9 locally and the issue was reproduced so I'll add it.

nicrowe00:
Checking .NET 6 causes everything to fail. I'm going to set it to specifically .NET 8.

sha1-validation/test.json
```diff
@@ -0,0 +1,18 @@
...
+    "type": "bash",
+    "cleanup": true,
+    "skipWhen": [
+    ],
+    "ignoredRIDs":[
+        "alpine",
+        "fedora",
+        "centos",
+        "rhel7",
+        "rhel8"
```
omajid, line 16: Taking a step back, the original customer report was that `HttpClientHandler` + `CheckCertificateRevocationList` fails against https://redhat.com. They told us that it works on RHEL 8 but not 9, which is why we are looking at a fix in RHEL 9. It seems to me that we want this test to run on all these platforms too, to verify that those platforms are not affected by the same issue that we confirmed exists on RHEL 9.

nicrowe00: I'll amend it.

sha1-validation/tests.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
omajid, line 5:
```suggestion
    <TargetFramework>$(TestTargetFramework)</TargetFramework>
```

omajid:
> I tested .NET 6 on UBI9 locally and the issue was reproduced so I'll add it.

CI doesn't reproduce the issue on .NET 6 on UBI 9 :confused:

nicrowe00:
> > I tested .NET 6 on UBI9 locally and the issue was reproduced so I'll add it.
> 
> CI doesn't reproduce the issue on .NET 6 on UBI 9 😕

Yeah, unfortunately I tested it when the target framework was set to .NET 8.

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+};
+
+HttpClient client = new HttpClient(handler);
+
+try
+{
+    HttpResponseMessage response = await client.GetAsync("https://redhat.com");
+
+    string responseBody = await response.Content.ReadAsStringAsync();
+    Console.WriteLine($"{sha1RsaSignatureOnLastElementInChain}");
```
tmds, line 18: We should fail when this is false with a meaningful message, something like:
```cs
if (!sha1RsaSignatureOnLastElementInChain)
{
    Console.WriteLine("The certificate chain that is validated is no longer using an RSA1 signature.");
    Console.WriteLine("The test must be updated to a different host.");
    return 1;
}
```

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    Console.WriteLine("PASS");
+    return 0;
+}
+catch (HttpRequestException e)
```
tmds, line 22:
```suggestion
catch (Exception e)
```

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    Console.WriteLine("\nException Caught!");
+    Console.WriteLine("Message :{0} ", e.Message);
```
tmds, line 25:
```suggestion
    Console.WriteLine(e);
```

This will print out the full exception including the stacktrace.

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Security.Cryptography.X509Certificates;
+
+bool sha1RsaSignatureOnLastElementInChain = false;
+
+HttpClientHandler handler = new HttpClientHandler { 
+    CheckCertificateRevocationList = true,
+    ServerCertificateCustomValidationCallback = ServerCertificateCustomValidation, 
+};
+
+HttpClient client = new HttpClient(handler);
```
tmds, line 11:
```suggestion
using HttpClient client = new HttpClient(handler);
```

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+catch (HttpRequestException e)
+{
+    Console.WriteLine("\nException Caught!");
+    Console.WriteLine("Message :{0} ", e.Message);
+    Console.WriteLine("FAIL");
+    return 1;
+}
+
+handler.Dispose();
+client.Dispose();
```
tmds, line 31:
```suggestion
```

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    return 0;
+}
+catch (HttpRequestException e)
+{
+    Console.WriteLine("\nException Caught!");
+    Console.WriteLine("Message :{0} ", e.Message);
+    Console.WriteLine("FAIL");
+    return 1;
+}
+
```
tmds, line 29:
```suggestion
}

Console.WriteLine("FAIL");
return 1;
```

sha1-validation/test.json
```diff
@@ -0,0 +1,15 @@
...
+    "name": "sha1-validation",
+    "enabled": true,
+    "requiresSdk": true,
+    "version": "6.0",
+    "versionSpecific": false,
+    "type": "bash",
+    "cleanup": true,
+    "skipWhen": [
+      "os=rhel,version=7", // We don't care about this test failing on .NET 7: https://issues.redhat.com/browse/RHEL-25254
+      "os=centos,version=7", // We don't care about this test failing on .NET 7: https://issues.redhat.com/browse/RHEL-25254
```
tmds, line 11:
```suggestion
      "os=rhel.9,version=7", // Test fails on RHEL9 with .NET 7: https://issues.redhat.com/browse/RHEL-25254
      "os=centos.9,version=7", // Test fails on RHEL9 with .NET 7: https://issues.redhat.com/browse/RHEL-25254
```

tmds:
@nicrowe00 https://github.com/redhat-developer/dotnet-regular-tests/pull/347/files#r1515533814 should be addressed, the other comments are some minor suggestions.

sha1-validation/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+    Console.WriteLine("\nException Caught!");
+    Console.WriteLine(e);
+}
+
+Console.WriteLine("FAIL");
+return 1;
+
+handler.Dispose();
+
+client.Dispose();
```
tmds, line 33: These lines can be removed (thanks to the `using`).

sha1-validation/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        var cert = element.Certificate;
+        Console.WriteLine($"{cert.SubjectName.Name} {cert.SignatureAlgorithm.FriendlyName}");
+    }
+    if ( chain.ChainElements.Last().Certificate.SignatureAlgorithm.FriendlyName != "sha1RSA" )
+    {
+        Console.WriteLine("The certificate chain that is validated is no longer using an RSA1 signature.");
+        Console.WriteLine("The test must be updated to use a different host.");
+        return false;
+    }
```
tmds, line 47: There may be multiple chains due to redirects.
The test is good is at least one of them uses a sha1 signature on the last cert.

By moving this check here, the test now requires all chains to meet that expectation.

This should go back to setting `sha1RsaSignatureOnLastElementInChain`, and we should check `sha1RsaSignatureOnLastElementInChain` is set after the HTTP response was read.

tmds: This is the reason the test is now failing on all CI jobs.

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+};
+
+using HttpClient client = new HttpClient(handler);
+
+try
+{
+    HttpResponseMessage response = await client.GetAsync("https://redhat.com");
+
+    string responseBody = await response.Content.ReadAsStringAsync();
+    Console.WriteLine($"{sha1RsaSignatureOnLastElementInChain}");
```
tmds, line 18:
```suggestion
    if ( chain.ChainElements.Last().Certificate.SignatureAlgorithm.FriendlyName != "sha1RSA" )
    {
        Console.WriteLine("FAIL");
        Console.WriteLine("The certificate chain that is validated is no longer using an RSA1 signature.");
        Console.WriteLine("The test must be updated to use a different host.");
        return 1;
    }
```

sha1-validation/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    foreach (var element in chain!.ChainElements)
+    {
+        var cert = element.Certificate;
+        Console.WriteLine($"{cert.SubjectName.Name} {cert.SignatureAlgorithm.FriendlyName}");
+    }
+    if ( chain.ChainElements.Last().Certificate.SignatureAlgorithm.FriendlyName != "sha1RSA" )
+    {
+        sha1RsaSignatureOnLastElementInChain = true;
+        Console.WriteLine("The certificate chain that is validated is no longer using an RSA1 signature.");
+        Console.WriteLine("The test must be updated to use a different host.");
```
tmds, line 42:
```suggestion
```

tmds (approved):

PR #349, opened 2024-05-29
Remove .NET 7 and F38

nicrowe00:
Both .NET 7 and F38 are EOL so they should be removed from the test matrix.


omajid (approved):
LGTM! Thanks!

PR #350, opened 2024-06-04
Removing .NET 7 from tests, removing outdated tests and increasing timeout-minutes

nicrowe00:
Cleaning up tests by removing references to .NET 7 and removing tests
made for older, now unsupported versions of .NET. Also increasing
the timeout limit for install .NET and test dependencies due to frequent
timeouts on Fedora.


managed-symbols-available/test.json
```diff
@@ -8,8 +8,7 @@
   "cleanup": true,
   "skipWhen": [
     "vmr-ci,version=6", // unpacking symbols is a hassle
-    "vmr-ci,version=7",
-    "non-packaged",
+    "non-packaged"
```
omajid, line 11: FYI: we often leave the trailing comma - even though that is not allowed (along with the `//` comments) by strict JSON implementations - to make it easier to add items to the list without more diff noise.

debugging-via-dotnet-dump/test.sh
```diff
@@ -319,18 +319,6 @@ grep 'Attributes: *PEFile' dump.out
...
-# dump-analyze 'name2ee *!System.String' > dump.out
-# cat dump.out
-# grep 'MethodTable:' dump.out | awk '{print $2}'
-# string_method_table=$(grep 'MethodTable:' dump.out | awk '{print $2}')
-# dump-analyze "dumpmt ${string_method_table}" > dump.out
-# cat dump.out
-# grep -E '^Name:[ \n\t]+System.String' dump.out
-# grep -E "^File:[ \n\t]+${framework_dir}" dump.out
-
-
```
omajid, line 333: Any chance we can review and then enable this for 8.0 and later?

managed-symbols-available/test.json
nicrowe00: Noted, thanks!

debugging-via-dotnet-dump/test.sh
nicrowe00: I enabled it and tested for .NET 6 and 8, seems to be working correctly.

omajid: Okay, this seems to be failing. My slight preference is to leave it commented out (like it was before this PR) so we can investigate it separately in the future. But I am okay if you want to remove it too.

omajid:
I opened https://github.com/redhat-developer/dotnet-regular-tests/pull/351 to remove the failing centos stream 8 legs. CentOS Stream 8 is EOL: https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/

omajid (approved):
Thank you!!

PR #353, opened 2024-06-11
skip tests run as "non-packaged" (non-rpm) validation

Vishwanatha-HD:
Added changes to skip the dotnet regular tests which are not required to run during cross build SDK validation.

### DIFF
--- a/bash-completion/test.json
+++ b/bash-completion/test.json
@@ -7,7 +7,8 @@
   "type": "bash",
   "cleanup": true,
   "skipWhen": [
-    "vmr-ci" // bash completion script not installed
+    "vmr-ci", // bash completion script not installed
+    "non-packaged",
   ],
   "ignoredRIDs":[
   ]
--- a/openssl-alpn/test.json
+++ b/openssl-alpn/test.json
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "non-packaged",
+  ],
   "ignoredRIDs":[
   ]
 }
--- a/publish-aspnet-selfcontained/test.json
+++ b/publish-aspnet-selfcontained/test.json
@@ -5,6 +5,9 @@
   "version": "8.0",
   "versionSpecific": false,
   "type": "bash",
-  "cleanup": true
+  "cleanup": true,
+  "skipWhen": [
+    "non-packaged",
+  ],
 }
 
--- a/publish-dotnet-selfcontained/test.json
+++ b/publish-dotnet-selfcontained/test.json
@@ -5,6 +5,9 @@
   "version": "8.0",
   "versionSpecific": false,
   "type": "bash",
-  "cleanup": true
+  "cleanup": true,
+  "skipWhen": [
+    "non-packaged",
+  ],
 }
 
--- a/restore-with-rid/test.json
+++ b/restore-with-rid/test.json
@@ -8,6 +8,7 @@
   "cleanup": true,
   "skipWhen":[
     "version=6,arch=s390x", // no runtime pack available
+    "non-packaged",
   ],
   "ignoredRIDs":[
   ]
--- a/sha1-validation/test.json
+++ b/sha1-validation/test.json
@@ -7,6 +7,7 @@
     "type": "bash",
     "cleanup": true,
     "skipWhen": [
+      "non-packaged",
     ],
     "ignoredRIDs":[
     ]
--- a/system-openssl/test.json
+++ b/system-openssl/test.json
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "non-packaged",
+  ],
   "ignoredRIDs":[
   ]
 }
--- a/file-permissions/test.json
+++ b/file-permissions/test.json
@@ -7,7 +7,8 @@
   "type": "bash",
   "cleanup": true,
   "skipWhen": [
-    "vmr-ci" // not packaged (tar.gz build)
+    "vmr-ci", // not packaged (tar.gz build)
+    "non-packaged",
   ],
   "ignoredRIDs":[
 
--- a/man-pages/test.json
+++ b/man-pages/test.json
@@ -7,7 +7,8 @@
   "type": "bash",
   "cleanup": true,
   "skipWhen": [
-    "vmr-ci" // man page not installed
+    "vmr-ci", // man page not installed
+    "non-packaged",
   ],
   "ignoredRIDs":[
   ]
--- a/rsa-pkcs-openssl/test.json
+++ b/rsa-pkcs-openssl/test.json
@@ -9,7 +9,8 @@
   "skipWhen": [
     "vmr-ci",           // upstream opts out of the OpenSSL change.
     "os=rhel.7",        // RHEL 7 is not getting OpenSSL changes.
-    "os=alpine"         // test validates behavior for Fedora/RHEL.
+    "os=alpine",         // test validates behavior for Fedora/RHEL.
+    "non-packaged",
   ],
   "ignoredRIDs":[
   ]
--- a/tools-in-path/test.json
+++ b/tools-in-path/test.json
@@ -6,7 +6,8 @@
   "type": "bash",
   "cleanup": true,
   "skipWhen": [
-    "vmr-ci" // tools PATH not configured (tar.gz build)
+    "vmr-ci", // tools PATH not configured (tar.gz build)
+    "non-packaged",
   ],
   "ignoredRIDs": [
   ]


Vishwanatha-HD:
@omajid. please review my changes and merge it, if the changes are fine.

bash-completion/test.json
```diff
@@ -7,7 +7,8 @@
   "type": "bash",
   "cleanup": true,
   "skipWhen": [
-    "vmr-ci" // bash completion script not installed
+    "vmr-ci", // bash completion script not installed
+    "non-packaged",
```
omajid, line 11: Can you add a comment to this line (sort of like what is in the previous line) to indicate why we are skipping this? Same goes for other json files.

Vishwanatha-HD: @omajid, I have added the comment for all the files to specify the reason for skipping the test. Please review the same. thanks..

openssl-alpn/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "non-packaged", // skip during cross-bld-sdk/tarball validation
```
tmds, line 10: How does this test fail?

publish-aspnet-selfcontained/test.json
```diff
@@ -5,6 +5,9 @@
   "version": "8.0",
   "versionSpecific": false,
   "type": "bash",
-  "cleanup": true
+  "cleanup": true,
+  "skipWhen": [
+    "non-packaged", // skip during cross-bld-sdk/tarball validation
```
tmds, line 10: How does this test fail?

publish-dotnet-selfcontained/test.json
```diff
@@ -5,6 +5,9 @@
   "version": "8.0",
   "versionSpecific": false,
   "type": "bash",
-  "cleanup": true
+  "cleanup": true,
+  "skipWhen": [
+    "non-packaged", // skip during cross-bld-sdk/tarball validation
```
tmds, line 10: How does this test fail?

restore-with-rid/test.json
```diff
@@ -8,6 +8,7 @@
   "cleanup": true,
   "skipWhen":[
     "version=6,arch=s390x", // no runtime pack available
+    "non-packaged", // skip during cross-bld-sdk/tarball validation
```
tmds, line 11: How does this test fail?

sha1-validation/test.json
```diff
@@ -7,6 +7,7 @@
     "type": "bash",
     "cleanup": true,
     "skipWhen": [
+      "non-packaged", // skip during cross-bld-sdk/tarball validation
```
tmds, line 10: This test is checking if we're applying a certain patch for RHEL9.
I wonder why this test does not fail on hydra's vmr builds (since they don't include the patch) :thinking: 
We should take a closer look.

cc @nicrowe00

system-openssl/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "non-packaged", // skip during cross-bld-sdk/tarball validation
```
tmds, line 10: How does this test fail?

Vishwanatha-HD:
Hi @tmds.. 
Omair and me had a discussion about the failures and please find his explanation below on why the tests fail when we try to validate the cross-build SDKs.. We mean "non-packaged" when we try to validate just the tarball and not the complete RPMs..

The errors fall into a few buckets:

Failing tests >>>>
  -> publish-aspnet-selfcontained
  -> publish-dotnet-selfcontained
  -> restore-with-rid
  -> openssl-alpn
  -> system-openssl:

- The above tests fail because they expect to be testing a source-built SDK and a cross-built SDK is different enough
  - For example, doesn't know about rhel.9-s390x RID
  - The cross-built SDK is portable, and doesn't link to some system libraries that the tests want to poke through
 
 
Failing tests >>>>
  -> bash-completion
  -> man-pages
  -> tools-in-path

- The above tests fail because they expect to be testing a packaged SDK, which has features like bash-completion enabled, which isn't a part of the .NET SDK tarball.

 
Failing tests >>>>
  -> file-permissions
  -> rsa-pkcs-openssl
  -> sha1-validation

- The above tests fail because they are testing Red Hat-specific changes, such as whether
  https://github.com/omajid/dotnet-runtime/commit/076687f5f9e7e1fce24f33f498b4e03c4150108e
  has been applied to the SDK.

Please let me know if you need more info and we will able to provide them.. Thanks..

tmds:
> why the tests fail when we try to validate the cross-build SDKs.. 

fyi, in hydra CI we don't run the regular tests against the cross-built SDKs. The cross-built SDK jobs provide SDKs for native builds. And those builds run the tests.

> For example, doesn't know about rhel.9-s390x RID

The cross-built SDK for .NET 9 gets built with a rid of `linux-390x` and provides assets under that rid. Rather than skipping these tests, they can run if they are using the appropriate SDK rid.

> The cross-built SDK is portable, and doesn't link to some system libraries that the tests want to poke through

Can you be explicit about this in the comments? I don't think `non-packaged` conveys these tests are skipped because the build is portable. Perhaps you could use something else, like `skipWhen: portable`.

> Failing tests >>>>

For these other tests, which should be ones that already have `vmr-ci` but where we are now adding now for the same reason: `non-packaged`.

It would be nice if we can avoid this duplication by either using `vmr-ci` or `non-packaged`, but not add them both.

For the `managed-symbols-available` test you can either also extract the symbols tarball so the test would pass, or add a new dedicated trait like: `skipWhen: no-symbols`.

Note that the test runner can be given multiple traits to skip, so you can set it to skip test matching any of: `vmr-ci`, `portable`, and `no-symbols`.

Vishwanatha-HD:
@tmds and @omajid, 
I have addressed all the code review comments and made the corresponding changes.
Please take a look and merge the CL if its fine.. thanks

system-openssl/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "portable", // skip during cross-bld-sdk/tarball validation
```
omajid, line 10:
```suggestion
    "portable", // portable dlopen()s system openssl
```

sha1-validation/test.json
```diff
@@ -7,6 +7,7 @@
     "type": "bash",
     "cleanup": true,
     "skipWhen": [
+      "vmr-ci", // skip during cross-bld-sdk/tarball validation
```
omajid, line 10:
```suggestion
      "vmr-ci", // relies on RHEL/Fedora specific patches
```

restore-with-rid/test.json
```diff
@@ -8,6 +8,7 @@
   "cleanup": true,
   "skipWhen":[
     "version=6,arch=s390x", // no runtime pack available
+    "portable", // skip during cross-bld-sdk/tarball validation
```
omajid, line 11:
```suggestion
    "portable", // portable doesn't know about platform RIDs
```

publish-dotnet-selfcontained/test.json
```diff
@@ -5,6 +5,9 @@
   "version": "8.0",
   "versionSpecific": false,
   "type": "bash",
-  "cleanup": true
+  "cleanup": true,
+  "skipWhen": [
+    "portable", // skip during cross-bld-sdk/tarball validation
```
omajid, line 10:
```suggestion
    "portable", // portable doesn't know about platform RIDs
```

publish-aspnet-selfcontained/test.json
```diff
@@ -5,6 +5,9 @@
   "version": "8.0",
   "versionSpecific": false,
   "type": "bash",
-  "cleanup": true
+  "cleanup": true,
+  "skipWhen": [
+    "portable", // skip during cross-bld-sdk/tarball validation
```
omajid, line 10:
```suggestion
    "portable", // portable doesn't know about platform RIDs
```

openssl-alpn/test.json
```diff
@@ -6,6 +6,9 @@
   "versionSpecific": false,
   "type": "bash",
   "cleanup": true,
+  "skipWhen": [
+    "portable", // skip during cross-bld-sdk/tarball validation
```
omajid, line 10:
```suggestion
    "portable", // portable dlopen()s system openssl
```

managed-symbols-available/test.json
```diff
@@ -8,7 +8,7 @@
   "cleanup": true,
   "skipWhen": [
     "vmr-ci,version=6", // unpacking symbols is a hassle
-    "non-packaged",
+    "no-symbols",  //skip during cross-bld-sdk/tarball validation
```
omajid, line 11:
```suggestion
    "no-symbols",  // standard VMR builds don't include symbols, it's an extra/manual step
```

publish-aspnet-selfcontained/test.json
tmds: It would be better to run this test (by using the SDK rid) than to skip it.

omajid: That's a good idea!

The current test runs `./runtime-id` to get the non-portable RID. Is there some way to identify/use the SDK's RID?

tmds: `dotnet --info` prints it under `RID`. You can also get it from the MSBuild `NETCoreSdkRuntimeIdentifier` property.

Vishwanatha-HD:
@omajid, I took a look at your changes to run the testcase with SDK runtime-ids rather than the RPM ones.. The changes are looking good.. Thanks for doing this for me.. !!

PR #355, opened 2024-06-19
Update package dependencies to latest stable versions.

tmds:
The current version we have for the test framework references vulnerable versions for System.Net.Http and System.Text.RegularExpressions which causes the test projects to fail to build with the latest .NET 9 SDK.

This updates all dependencies to their latest stable versions.

cc @nicrowe00 @omajid


tmds:
CI failures are unrelated to the changes.

omajid:
Vulnerable versions fail to restore with .NET 9 SDK? That's a pretty cool feature!

PR #359, opened 2024-06-24
Add a test for non-portable SDK RID

omajid:
We have recently updated a number of tests to use the SDK's RID (instead of the expected OS non-portable RID). We should have a test that explicitly tests that the SDK's RID matches the expected OS-specific non-portable RID. This is that test.

### DIFF
--- a/non-portable-rid/test.sh
+++ b/non-portable-rid/test.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS=$'\n\t'
+set -x
+
+sdk_rid="$(../runtime-id --sdk)"
+os_rid="$(../runtime-id)"
+
+if [[ $sdk_rid != $os_rid ]]; then
+    echo "SDK RID ($sdk_rid) did not match expected RID ($os_rid)"
+    exit 1
+fi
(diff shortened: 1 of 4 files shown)


non-portable-rid/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS=$'\n\t'
+set -x
+
+sdk_rid="$(../runtime-id --sdk)"
+os_rid="$(../runtime-id)"
```
tmds, line 8: I assume this means we're no longer calling this script without an argument anywhere else but here?

tmds (approved):
LGTM.

non-portable-rid/test.sh
tmds: For the following, we expect it to be the SDK id, so we might as well pass `--sdk` in the tests:
```
distribution-packages/test.sh
10:runtime_id=$(../runtime-id)

publish-ready-to-run/test.sh
8:dotnet publish -r "$(../runtime-id)" /p:PublishReadyToRun=true
```

For the others, we're not actually using it as a rid but to identify the distro.

omajid: I am not sure about `distribution-packages`, but I fixed up `publish-ready-to-run` too.

tmds: > I am not sure about distribution-packages

I only see this usage for the rid:

https://github.com/redhat-developer/dotnet-regular-tests/blob/c3b481c4f22e55cc563d78e81e99fbedfd829a39/distribution-packages/test-standard-packages#L154

omajid: Right. I am just not sure we want the test to pass if `/usr/lib64/dotnet/packs/Microsoft.NETCore.App.Host.linux-x64/` (for example) is present or not.

tmds: The test you are adding here is meant to verify the SDK rid is the expected non-portable rid.

Other tests that fail for the same reason just add noise to diagnose the problem.

omajid: Okay, that's fair.

PR #362, opened 2024-08-19
Add test for source-built NativeAOT.

tmds:


tmds:
The test added here is failing on the fedora:40 image with .NET 9 preview 7 due to missing dependencies when linking:

```
  console -> /__w/dotnet-regular-tests/dotnet-regular-tests/nativeaot-sb/bin/Release/net9.0/fedora.40-x64/console.dll
  Generating native code
  /usr/bin/ld.bfd: cannot find -lssl: No such file or directory
  /usr/bin/ld.bfd: cannot find -lcrypto: No such file or directory
```

To avoid the user having to figure out what packages to install, we can require these packages from our `dotnet-sdk-aot` package. @omajid wdyt?

tmds:
> I couldn't agree more. I already implemented this, but didn't test it fully. so I must have missed some stuff: https://gitlab.com/redhat/centos-stream/rpms/dotnet9.0/-/blob/c10s/dotnet9.0.spec?ref_type=heads#L413-414

You can use this test to check.

@omajid, do you have feedback about the PR, or is it good to merge?

nativeaot-sb/test.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+IFS=$'\n\t'
+set -x
+
+RID_ARG="/p:RuntimeIdentifier=$(../runtime-id --sdk)"
+
+for DEPENDENCY in NO_DEPS DEP_CRYPTO DEP_ZLIB DEP_BROTLI
```
omajid, line 9: Other dependencies (eg, `libunwind`, `rapidjson`) are mandatory for the runtime (if the runtime uses the system version of those libraries). So the program will probably fail to start if they are missing. Should we document those here?

tmds: I've added a comment.

PR #363, opened 2024-09-12
Updating test matrix

nicrowe00:


.github/workflows/ci.yml
```diff
@@ -49,19 +56,23 @@ jobs:
         run: |
           set -euo pipefail
           cat /etc/os-release
-          if grep fedora /etc/os-release ; then
-            if [[ ${{ matrix.dotnet_version }} == 9.* ]]; then
-              dnf install 'dnf-command(copr)' -y
+          if [[ ${{ matrix.dotnet_version }} == 9.* ]]; then
```
tmds, line 59: When reviewing this there are some things the reviewer needs to know:
- .NET 9 is a preview, that's why it needs copr
- .NET 9 introduced the sdk-aot package
- /etc/os-release is going to match the grep fedora also on Fedora derived OSes (like CentOS)
- .NET 8 introduced the dbg packages

Perhaps we can include this information with some comments.

tmds:
@nicrowe00 feel free to merge if/when the CI is happy with the changes. Thanks!

.github/workflows/ci.yml
nicrowe00: Thanks for the suggestion, I'll add it now.

omajid (approved):
Thanks!

PR #367, opened 2025-01-22
Eliminate dependency on .NET 10's non source-buildable Microsoft.AspNetCore.App.Internal.Assets package.

tmds:
This enables running tests against .NET 10 build which are currently failing due to https://github.com/redhat-developer/dotnet-regular-tests/issues/368/. Once the upstream issue is fixed, these changes can be reverted.

### DIFF
--- a/cgroup-limit/cgroup-limit.csproj
+++ b/cgroup-limit/cgroup-limit.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <OutputType>Exe</OutputType>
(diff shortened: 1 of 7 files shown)


omajid:
The tests run on 8.0 and 9.0 as well, will this have any impact on those? CI thinks they should be okay. If that's true, then okay by me.

omajid (approved):

tmds:
> The tests run on 8.0 and 9.0 as well, will this have any impact on those? CI thinks they should be okay. If that's true, then okay by me.

GitHub + internal CI will catch regressions introduced for 8.0 and 9.0.

Issue #369, opened 2025-01-23
Skip tests in GitHub CI using traits


tmds:
Some tests are being skipped by removing them:

https://github.com/redhat-developer/dotnet-regular-tests/blob/20a3d7214f1b0b7dd57e8de69768fcd352f53258/.github/workflows/ci.yml#L97-L116

It would be preferable to skip them using a trait. This gives visibility in the pull requests that these tests are being skipped.

cc @nicrowe00 @omajid

PR #372, opened 2025-02-24
Use traits in GitHub CI

omajid:
Instead of the hack of rm -rf'ing specific test directories in CI, use traits to skip them in CI.

Other changes:

- Alpine edge seems to be okay with RIDs now. No need to skip them.

- bash-completion and strace seem to be available in UBI repos

Fixes: #369

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,28 +94,18 @@ jobs:
         run: |
           set -euo pipefail
 
-          ### This is very questionable, but making our CI fail because distro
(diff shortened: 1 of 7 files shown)


omajid:
Fedora rawhide has a known issue. It was recently renamed from Fedora 42 to Fedora 43, and the SDK hasn't learned about `fedora.43` yet. The next rebuild of the SDK in Fedora should sort that out. But until it does, this test will continue to fail.

tmds (approved):
lgtm, thanks!

PR #379, opened 2025-05-02
Update CI and add .NET 10 copr to test matrix

omajid:


omajid:
There are 4 failures in .NET 10 in copr:

-  assemblies-valid: https://github.com/dotnet/aspnetcore/issues/60174
- dotnet-monitor-works: Can't fix. there's no 10.0 version of the package on nuget.org
- host-probes-rid-assets-legacy: Will fix itself. Just needs a source-build of .NET 10 that is built against itself.
- man-pages: https://github.com/dotnet/sdk/pull/48784

nicrowe00 (approved):

PR #404, opened 2025-12-15
Log lldb version when using it

omajid:
I have seen an unexpected failure when using lldb to debug a core in a CI enviornment. Log additional details to help make it easier to reproduce/fix the problem.

### DIFF
--- a/debugging-sos-lldb-via-core/test.sh
+++ b/debugging-sos-lldb-via-core/test.sh
@@ -49,6 +49,7 @@ if ! command -v lldb ; then
     echo "lldb is not installed"
     exit 1
 fi
+{ rpm -qa | grep lldb; } || { apk list -i | grep lldb; } || true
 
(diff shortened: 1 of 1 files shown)


omajid:
Warning looks like this:


```
warning: (aarch64) /usr/lib64/dotnet/dotnet unsupported DW_FORM values: 0x1f20 0x1f21
```

debugging-sos-lldb-via-core/test.sh
```diff
@@ -49,6 +49,7 @@ if ! command -v lldb ; then
     echo "lldb is not installed"
     exit 1
 fi
+{ rpm -qa | grep lldb; } || { apk list -i | grep lldb; } || true
```
tmds, line 52: Perhaps you could implement https://github.com/redhat-developer/dotnet-bunny/pull/96#discussion_r2523523080 instead?

omajid: Done.

omajid:
Closing in favour of https://github.com/redhat-developer/dotnet-bunny/pull/97

PR #331, opened 2024-01-26
Disable tests that fail on armv7

ayakael:
A few test units fail on on armv7, which I suppose has to do with certain features not implemented on coreclr-arm:
* nativeaot: AOT not implemented on arm
* pgo-dynamic: pgo-dynamic not inplemented on arm

These I'm not sure if they are implemented on armv7:
* debugging-via-dotnet-dump
* debugging-sos-lldb-via-dump

PR #409, opened 2026-02-17
debugging-via-dotnet-dump: Remove "linux-arm" from ignoredRIDs

### LINKED ISSUE #236: Fix this test on aarch64
https://github.com/redhat-developer/dotnet-regular-tests/blob/66b6dc298622b8a890cff720fa87683481ef604f/debugging-via-dotnet-dump/test.sh#LL214C26-L214C26

This fails on aarch64 for some reason

nicrowe00:
Removing the "linux-arm" RID from the ignoredRIDs section for debugging-via-dotnet-dump as the test passes on arm. Also the rid for arm is "linux-arm64" so this test was running on arm machines regardless.

Closes #236

### DIFF
--- a/debugging-via-dotnet-dump/test.json
+++ b/debugging-via-dotnet-dump/test.json
@@ -8,7 +8,6 @@
   "cleanup": true,
   "timeoutMultiplier": 2,
   "ignoredRIDs": [
-     "linux-arm" // dotnet-dump relies on features not implemented on arm
   ],
   "skipWhen": [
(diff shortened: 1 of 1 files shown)


omajid (changes requested):
We added this in https://github.com/redhat-developer/dotnet-regular-tests/pull/331 for arm (32) support.

nicrowe00:
OK, I'll go ahead and close this.

PR #411, opened 2026-04-20
Add tls-uses-pqc test to verify OpenSSL TLS does PQC key exchange.

tmds:


tls-uses-pqc/Program.cs
```diff
@@ -0,0 +1,73 @@
+using System.Net;
+using System.Net.Security;
+using System.Net.Sockets;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+// Create a self-signed certificate.
+using var key = ECDsa.Create();
```
omajid, line 13: Should we force a PQC algorithm here?

tmds: The test intentionally has zero PQC code and then check TLS used PQC.

omajid: Can you add a note about this somewhere?

tmds: I've added a comment to the top of `Program.cs`.

omajid (approved):
Failing tests are related to https://github.com/dotnet/source-build/issues/5531 and https://github.com/redhat-developer/dotnet-regular-tests/pull/412

PR #412, opened 2026-04-22
ci: add rpm depend on Alpine

ayakael:


omajid:
Do you know where tests are trying to use rpm on alpine? Does installing the `rpm` package make those tests work correctly on alpine?

ayakael:
Nothing has changed downstream on our side. I don't understand why `rpm` is now expected.

omajid:
Ah, it's https://github.com/redhat-developer/dotnet-bunny/blob/6ce0abfa1285a803ed0d055b557afb1bb0debeed/Turkey/TestRunner.cs#L182-L201 which should be handling rpm vs apk correctly, but doesn't seem to.

PR #413, opened 2026-05-11
debugging-via-dotnet-dump: remove WeakLong check.

### LINKED ISSUE #407: debugging-via-dotnet-dump fails with "too few references of type WeakLong"
This test is sometimes failing with:

```
++ grep -cF WeakLong
+ [[ 0 -lt 2 ]]
+ echo 'fail: too few references of type WeakLong'
+ exit 2
fail: too few references of type WeakLong
```

cc @omajid

tmds:
WeakLong handles are not reliably present in every dump causing intermittent test failures.

The remaining checks still provide sufficient coverage for gchandles validation.

Fixes https://github.com/redhat-developer/dotnet-regular-tests/issues/407.

### DIFF
--- a/debugging-via-dotnet-dump/test.sh
+++ b/debugging-via-dotnet-dump/test.sh
@@ -429,7 +429,7 @@ cat dump.out
 heading "gchandles"
 dump-analyze 'gchandles' > dump.out
 cat dump.out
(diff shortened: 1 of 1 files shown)


omajid (approved):
Thanks!

PR #422, opened 2026-07-01
tls-uses-pqc: support running against OpenSSL 4.

tmds:


tls-uses-pqc/Program.cs
```diff
@@ -55,6 +57,23 @@
 return 1;
 
 // Use OpenSSL interop to get the negotiated key exchange group (SslStream doesn't expose this information).
```
omajid, line 59: Should this comment be moved now?

tmds: Everything below the comment is for the interop.

tls-uses-pqc/Program.cs
```diff
@@ -55,6 +57,23 @@
...
+    {
+        if (name == "libssl")
+        {
+            if (NativeLibrary.TryLoad("libssl.so.4", assembly, searchPath, out var h) ||
+                NativeLibrary.TryLoad("libssl.so.3", assembly, searchPath, out h))
+            {
+                return h;
+            }
+        }
+        return IntPtr.Zero;
```
omajid: I was confused at first the behaviour of `IntPtr.Zero` is. Maybe add a comment like:

```suggestion
        return IntPtr.Zero;  // fall back to default resolver
```

Or maybe the keyword `default` is both clearer and more confusing at the same thing?

```suggestion
        return default;
```

PR #425, opened 2026-07-14
Add mcpserver to ignored templates

drehak:
New template in .NET 10, but at the moment there's a vulnerability which causes it to fail tests:

https://github.com/advisories/GHSA-v5pm-xwqc-g5wc


omajid (approved):
Thanks!

PR #431, opened 2026-08-13
tools-in-path: treat Fedora 46 like 45

omajid:
It's a new Fedora version. The behaviour is the same.


.github/workflows/ci.yml
```diff
@@ -32,6 +33,15 @@ jobs:
           - "8.0"
           - "9.0"
           - "10.0"
+        exclude:
+          - container_image: registry.fedoraproject.org/fedora:45
```
omajid, line 37: With upcoming EOL, .NET 8 and 9 have been dropped from Fedora 45 and later.

nicrowe00 (approved):
LGTM. Thanks!

